PR #179, opened 2019-04-23
Fix some Material demos errors for GType

dmitrys99:


dmitrys99:
Partially cherry-picked, partially solved by #183 by fixing type verifier.
Will not merge.

PR #192, opened 2019-04-26
Add Warning when a path in includes does not exist

magnusklundgren:


alstrup:
We have the innovation path hardcoded in the flow compiler, but that does not make sense in the open source version. We would prefer not to have a warning for that one, and only complain about includes given by the user.

magnusklundgren:
Changed to only include paths specified by the -I parameter

alstrup:
Do we also check the includes from the flow.config file? That is important as well.

magnusklundgren:
Yes. we check it now

alstrup:
LGTM

PR #193, opened 2019-04-26
Cover RTL functionality by unit tests - made switching lang in runtime

artxp:


artxp:
https://github.com/area9innovation/innovation/pull/539

PR #220, opened 2019-05-06
Promise. Add `sequenceP`.

wowofbob:
`parallelP` analogue which computes supplied promises sequentially

### DIFF
--- a/lib/promise.flow
+++ b/lib/promise.flow
@@ -79,6 +79,11 @@ export {
 
 	//mapP and mapPparallel were moved to arrayutils.flow
 
+	// This function takes an array of promises and ties them together
+	// into one `bindP` expression. Meaning that all promises will be
+	// computed sequentially. Use it to avoid writing a `bindP` chain
+	// manually.
+	sequenceP(steps : [Promise<?, ??>]) -> Promise<[?], ??>;
 
 	// These could be useful when you do not have a common type, because it could allow you to construct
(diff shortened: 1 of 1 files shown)


alstrup:
@dmitrybure LGTM, but you wrote promises, so worth giving you a vote on it.

lib/promise.flow
```diff
@@ -79,6 +79,11 @@ export {
 
 	//mapP and mapPparallel were moved to arrayutils.flow
 
+	// This function takes an array of promises and ties them together
+	// into one `bindP` expression. Meaning that all promises will be
+	// computed sequentially. Use it to avoid writing a `bindP` chain
+	// manually.
```
dmitrybure: document what happens when there is an error

wowofbob: I noticed that there is `allP` function which does exactly the same as `sequenceP` but has a different implementation. It uses recursion and has optimization to keep a call stack from growing too much, while `sequenceP` uses `fold` from `array.flow`. I added `sequenceP` in order to save me some time of writing `bindP` chains. So I'm not sure which one is better and what to do about it. What do you think? @alstrup @dmitrybure

dmitrybure: if it's exactly the same, also in error handling, I would use allP

wowofbob:
Introducing similar to `allP` function is not necessary, closing this PR

PR #224, opened 2019-05-14
Small refactoring in MMultiAudioPlayer()

dmiserg:


lib/material/material_ui.flow
```diff
@@ -362,6 +362,13 @@ MAudioPlayer(filename : string, style : [MAudioPlayerStyle]) -> Material {
 	)
 }
 
+SampleStates(
```
pzotov: Why don't you use array of
SampleState(
	length : DynamicBehaviour<double>, 
	position : DynamicBehaviour<double>, 
	isPlay : DynamicBehaviour<bool>, 
	isReady : DynamicBehaviour<bool>
)?

dmiserg: Because there are some part like fconnect(fsum(states.length), lenEx)

PR #225, opened 2019-05-14
Jit overlap solution

Coolrunner86:
New one

### DIFF
--- a/platforms/common/cpp/core/JitProgram.cpp
+++ b/platforms/common/cpp/core/JitProgram.cpp
@@ -9,6 +9,8 @@
 using namespace asmjit;
 using namespace asmjit::x86;
 
+const unsigned long MAX_JIT_MEMORY = 256*1024*1024;
+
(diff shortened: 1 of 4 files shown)


platforms/common/cpp/core/JitProgram.cpp
```diff
@@ -376,7 +374,7 @@ bool FlowJitProgram::compile()
 
     // Initialize generation
     next_code_off = 0;
-    next_data_off = committed_data_off = MAX_CODE_MEMORY;
+	next_data_off = committed_data_off = std::max(memory_limit, (unsigned long)flow_code.size() * 25);
```
alstrup: Keep a 256mb minimum size, just to ensure we do not break existing code in some random circumstance. Then we are sure this is a monotoneous improvement.

Coolrunner86: See no point in this, since we had problems only with rhapsode, but ok. Done.

PR #227, opened 2019-05-16
Multiplayer: added opportunity to switch between players

dmiserg:
Added opportunity to select a player in MMultiAudioPlayer element.

### DIFF
--- a/lib/fform/fform.flow
+++ b/lib/fform/fform.flow
@@ -130,7 +130,7 @@ export {
 	// If widthheihgt is 0,0 size is automatic, otherwise defined
 	FVideo(filename : string, wh : DynamicBehaviour<WidthHeight>, style : [FVideoStyle]);
 		FVideoStyle ::= FVideoKeepAspectRatio, FVideoPlay, FVideoPosition, FVideoLength, FVideoVolume, FVideoFullscreen, FVideoLoop, FVideoControls, FVideoSubtitles,
-			FVideoTimeRange, FVideoPlaybackRate, FVideoRealSize, FVideoGetCurrentFrame, SynchroCalls, StreamStatus, OnVideoLoadingError, ViewBounds;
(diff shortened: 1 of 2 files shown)


lib/material/material_ui.flow
```diff
@@ -389,124 +389,136 @@ SampleStates(
...
- 				\-> subscribe(currIndexB, \index ->
- 					changeAudioPlaying(index, getValue(playEx))
- 				),
- 				\-> subscribe(playEx, \isPlay ->
- 					changeAudioPlaying(getValue(currIndexB), isPlay)
- 				)
- 			],
- 			MGroup(players)
- 		)
+		{
```
stanislavpodolia: No point in moving everything inside MShow since this code will be executed anyway despite of any const(i2b(amount)) value.
Also MShow(const(...), ...) makes no sense, plain if (...) {...} else { TEmpty(); } will do the trick and won't mislead anyone but it's a pretension to the one who wrote it first not to you.

lib/fform/fform.flow
```diff
@@ -152,6 +152,8 @@ export {
 			FVideoRealSize(size : DynamicBehaviour<WidthHeight>);
 			// get current frame as base64 string
 			FVideoGetCurrentFrame(fn : ref () -> string);
+			// to switch files (samples) in multiplayer
+			IndexInPlaylist(index : DynamicBehaviour<int>);
```
stanislavpodolia: Missed this at first glance, IndexInPlaylist should go to different union, it means nothing in FVideo case.
Add something like MMultiAudioPlayerStyle ::= FVideoStyle, IndexInPlaylist to material_ui.flow, move IndexInPlaylist there and change mstyle : [FVideoStyle] to mstyle : [MMultiAudioPlayerStyle] in MMultiAudioPlayer.
In case you need to pass mstyle to TVideo, videoPlayerStyle2FVideoStyle in material_player.flow is an example of extracting FVideoStyle from other union.

lib/material/material_ui.flow
dmiserg: Sure, thanks. Fixed.

lib/fform/fform.flow
dmiserg: Well, I got it.

PR #231, opened 2019-05-21
More iterators for cpp target (needed for 518_word-per-char-metrics-native)

brawaga:


brawaga:
Analyzed via valgrind, looks master and this branch shows equivalent non-growing momory leak about 20k.

vereschagai:
Looks good

PR #232, opened 2019-05-21
fixed `requestExceptionDb` function

Eugeny48:
Failures that occur in conjunction with an individual query are reported by `QSqlQuery::lastError()`, not by `QSqlDatabase::lastError()` (https://doc.qt.io/qt-5/qsqldatabase.html#lastError)

Unfortunately I can not build x64 version


alstrup:
LGTM. Would be great to have this fix

PR #239, opened 2019-05-26
Flowc incremental java build 1

dmitry-vlasov:


alstrup:
Seems there are two separate changes. Do they go together?

PR #241, opened 2019-05-27
Use absolute path to default fontconfig in resources

hey-sancho:
This allows to use the default fontconfig from `resources/`, even when it's not specified in flow.config.

I'm not sure if I should recompile any binaries, so would be glad for some tips. Thanks!

### DIFF
--- a/tools/common/compilerconfig.flow
+++ b/tools/common/compilerconfig.flow
@@ -111,7 +111,7 @@ getCompilerConfig() {
     partialevaluate = getUrlParameterDef("pe", "");
     prettyprint = getUrlParameterDef("pp", "");
 
-    fontconfig = getUrlParameterDef("fontconfig-file", "");
+    fontconfigFile = getUrlParameterDef("fontconfig-file", "");
     dumpids = getUrlParameter("dump-ids");
 
     htmlMinify = isUrlParameterTrue("html-minify");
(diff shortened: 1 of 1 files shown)


tools/common/compilerconfig.flow
```diff
@@ -241,7 +241,7 @@ getCompilerConfig() {
                 beautify,
                 verbose,
                 nwjsmode,
-                if (fc == "") "resources/fontconfig.json" else fc,
+                if (fc == "") flowDir + "resources/fontconfig.json" else fc,
```
alstrup: I think you should first use "fileExists" on the relative path, and if that exists, use that. If that does not exist, then OK to use the flowDir path. We do want to support relative paths.

tools/common/compilerconfig.flow
```diff
@@ -209,14 +209,20 @@ getCompilerConfig() {
...
+            Pair("fontconfig-file", fontconfigFile),
             Pair("js-call-main", jsCallMain),
             Pair("force-build", fcForceBuild),
             Pair("incremental", if (fcIncremental) "1" else ""),
             Pair("js-no-otc", noOtc),
             Pair("server-port", fcServerPort),
         ]), configFile);
-        fc = getConfigParameterDef(configFile, "fontconfig-file", fontconfig);
+
+        // FontConfig
```
alstrup: Probably worth adding a comment explaining that it falls back to the flowdir if the given path interpreted as relative does not exist. OK with that.

PR #247, opened 2019-05-31
pebble : hierarchy

shkel:
https://trello.com/c/XOSbbnqj/2918-new-pebble-design

https://git.area9lyceum.com/Lyceum/lyceum/compare/master...curator_hierarchy


dmitrybure:
no changes in common pebbles code, only hierarchy which is only used in lyceum, so if it works, fine. 
as I can see it changes the way we treat items parameters, but I guess that's what was required 
LGTM

PR #251, opened 2019-06-05
Promise. Introduce 'eitherP'.

wowofbob:
Add function to construct `Promise` which can fallback on error.

@yegordovganich 
@dmitrybure


dmitrybure:
code is ok, but is it worth adding global function like this? do you have at least three places in the code where this pattern is used?

wowofbob:
I have it only in one place so far. I can postpone the merge but continue to use a global name like this. When it will come up in some other places or in my code again, I will re-open this PR. What do you think? @dmitrybure

dmitrybure:
yes, good idea. we don't want to "pollute" global places like this too much, as it will complicate understanding of the API for others

PR #260, opened 2019-06-14
Flowc delete dir deletes recursively

dmitry-vlasov:


alstrup:
What does c++ do?

dmitry-vlasov:
> What does c++ do?

Hm, c++ calls rmdir, which doesn't remove directory recursively. But this function is convenient, on my point of view. Maybe I should add deleteDirectoryRecursively native for it?

alstrup:
Why native? Can be implemented in flow and might already exist.

fre. 14. jun. 2019 07.05 skrev Dmitry Vlasov <notifications@github.com>:

PR #262, opened 2019-06-14
wigi: we should track detached words for cursor related stuff

ykuzmenko73:
In edit mode, wigi traverse interactive elements (IE) list made from paragraph in order to find the real position of the cursor on the screen. but reflow procedure removes last space on the each line that does not fit the width so such IE does not recieve updates of it's coordinates and the positionin of the cursor gets wrong.
We can't just keep such spaces because it would lead to horizontal scroll appear and we can't put them on the next line because it will break align.
So we have to introduce attach/detach property for the words.

### DIFF
--- a/lib/form/paragraph.flow
+++ b/lib/form/paragraph.flow
@@ -45,7 +45,7 @@ export {
 			TextFragmentMoreLineBreaks();
 			TextFragmentLastChild();
 			// Index of current paragraph element the fragment belongs to.
-			// Needed to handle features that belong to whole paragraph element. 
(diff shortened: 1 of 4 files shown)


ykuzmenko73:
[arrayDiff: add new ArraySwap operation](https://trello.com/c/qkdVisXe/591-arraydiff-add-new-arrayswap-operation)

ykuzmenko73:
[RTE bug with cursor movement in Curator](https://trello.com/c/LiyT5r2B/5255-rte-bug-with-cursor-movement-in-curator)

PR #265, opened 2019-06-17
Check special chars in filename

DenisPushkarev:


lib/material/internal/filebrowser/material_filebrowser_internal.flow
```diff
@@ -15,6 +15,9 @@ export {
 
 	//checks if partition name has no conflicts with existing names
 	validatePartitionName(name : string, errorB : DynamicBehaviour<Maybe<Pair<string, bool>>>, existsNames : [string]) -> void;
+
+	// Symbols, forbidded in the names
+	forbiddenChars : [string] = ["\\", "/", ":", "*", "?", "<", ">", "|", "\"", "\n", "\t"];
```
andrejefimov: Move it back to filesystemdb, it is specific for that component, and this file contains a general representation for all file browsers. Pass it as a parameter to validatePartitionName and validateFileName

PR #278, opened 2019-06-26
requestDb: fix double query execution

Ihor0k:


platforms/qt/qt-backend/DatabaseSupport.cpp
```diff
@@ -158,7 +158,7 @@ StackSlot DatabaseConnection::requestDb(RUNNER_ARGS) {
     RUNNER_CheckTag(TString, rawquery);
 
     QString queryString = RUNNER->GetQString(rawquery);
-    QSqlQuery *query = new QSqlQuery(queryString, db);
```
Ihor0k: This QSqlQuery constructor executes queryString automatically

dmitrybure: how fun

PR #280, opened 2019-06-27
We shouldn't terminate application while some deffered code is running

dimabure:
From javadoc: 

```
public Timer(boolean isDaemon)
Creates a new timer whose associated thread may be specified to run as a daemon. A daemon thread is called for if the timer will be used to schedule repeating "maintenance activities", which must be performed as long as the application is running, but should not prolong the lifetime of the application.
```

I don't know what is the benefit of having thread as daemon, but having it as daemon cause following code working incorrect (at least different from other targets):

```
main() {
  ...
  deffered(\-> {
  ...
     foo()
  })
}
```

In Java `foo` will never be executed (`deffered` implemented using `timer`)


dmitrybure:
History in github is short and I don't know who implemented this function like this originally

PR #286, opened 2019-07-01
Remove size inspecting from decorations forms, because size is unused

chekaninmaxim:
I have provided deep code investigation and noticed that size of decorated forms is not used anywhere.
There is only one place where it's used and it's lead to bug of decreasing image sizes after zoomming. So, I have removed these sizes.
@ViktorKurach Could you please, check if your project is not broken with this branches?
https://github.com/area9innovation/innovation/pull/773

### DIFF
--- a/lib/ui/imagedecorated.flow
+++ b/lib/ui/imagedecorated.flow
@@ -21,7 +21,7 @@ export {
 		filename : string,
 		width : double,
 		height : double,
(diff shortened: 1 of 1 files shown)


chekaninmaxim:
[Captions for images makes images scale crazy](https://trello.com/c/UjzISTdX/5451-captions-for-images-makes-images-scale-crazy)

PR #290, opened 2019-07-04
Tagged videos shouldn't show time - ability to set range for video UI

cs1924:
https://trello.com/c/xw3ulHLq/4939-tagged-videos-shouldnt-show-time - Video player should display fragment of video just like separate video file.
Maybe there is better way to achieve the same result.


cs1924:
> Generally looks good, but wouldn`t it be better to set FVideoTimeRange(visibleRange.start, visibleRange.end) to inner TVideo as well?

No, this style causes problem described in the card. You can see only selected video fragment, but controls show that it is a part of video file. We need controls look like a whole file is played.

NikitaFil:
I mean to set FVideoTimeRange in addition to your changes for player, so new style will affect both real and visible video time.

cs1924:
Agree, that's a good idea.

PR #296, opened 2019-07-12
Added function arrayGroupBy

natkoma:


lib/ds/arrayutils.flow
```diff
@@ -99,6 +99,8 @@ export {
 	// Sorts as and groups the elements into equivalence classes (as given by
 	// compare). Equivalence classes are sorted according to compare.
 	sortGroupBy : (as : [?], compare : (?, ?)->int) -> [[?]];
+	// Groups the elements into equivalence classes (as given by compare).
```
yegordovganich:
```suggestion
	// Groups the elements into equivalent classes (as given by compare).
```

natkoma: Added example

PR #305, opened 2019-07-23
dceil and frac - use full values range, not only integers

anastasiakucherova:
When I see a function like fn(value : double) -> double, I expect it works actually with doubles and doesn't miss a huge range of data that's out of intMin - intMax.
What if we do the change, would it break something?


anastasiakucherova:
Yes, for negative values it works the same way for integers as before changes

PR #308, opened 2019-07-26
fixing a bug in `requestDbMulti`

Eugeny48:
allocating memory inside a loop and collecting into a vector leads to deallocating memory when we leave the loop.

UPD: It looks like the memory was freed by the flow's garbage collector. It's solved with native Arrays instead of vectors.


platforms/qt/qt-backend/DatabaseSupport.cpp
```diff
@@ -206,27 +211,40 @@ StackSlot DatabaseConnection::requestDbMulti(RUNNER_ARGS) {
         DatabaseResult *databaseResult = new DatabaseResult(this, query);
         query->exec(queryString);
 
-        last_error = query->lastError().text().trimmed();
+        QString err_msg = query->lastError().text().trimmed();
+        if (err_msg.length()) {
+            last_error = err_msg;
+        }
 
         std::vector<StackSlot> resultSets;
```
Ihor0k: This variable is not used anymore

Eugeny48: Thanks

PR #309, opened 2019-07-26
Translation wigi light api support

kszh:
Added new wigi light api support for translation with "g" and "x" tags. These tags must be unescaped in xliff/json files, but text inside tags must be escaped.

Also these changes suppose that wigi2xliff_light and innovation/components/translationtool/... used only for Lyceum project. Vladimir, please, correct me if I am mistaken.

TODO: update repo of translation tool on tech server
TODO: update Lyceum repo
TODO: write and apply fixup DB for existing Rhapsode translations
TODO: check curator translation tool and email editor (Lyceum repo)
TODO: refactor innovation/components/translationtool to use translations api

@ViktorKurach @pzotov 
FYI


ViktorKurach:
@kszh thank you for your work! Please notify me when you are merging the PR here and in innovation

lib/text/translationutils.flow
```diff
@@ -1,11 +1,9 @@
-import string;
 import text/languages;
-import text/translation_fixup;
-import text/translationapi; // ContentUnit
 import fs/filesystem; // fileNameOnly
 import formats/json/json;
-import math/stringmath;
+import text/translationapi; // ContentUnit
 import ui/fontmapping; // getLang
+import formats/xml;
```
vladimirgorky: We minimized imports and excluded xml stuff for simple strings. It is why there was simplified xmlEscape function.
Lightweight application itself don't  need xml for translation - it only loads json with simple strings.
To keep this approach I think we need to restore this here and exclude xml import from this utils file.

alstrup: This small amount is OK to have in flow9.

vladimirgorky: Import xml - OK if no problem with app file size now. But my question was not about this import but about functions like this:

xliffNode2WigiLightString(node : XmlNode, level : int, maybeParent : Maybe<XmlNode>) -> string;

This is utilities for wigi format but flow9 as open source knows nothing about wigi so it seems to me it should be placed in innovation repo like translationwikisupport.flow.

kszh:
merged

PR #315, opened 2019-08-06
snackbar width by text width

Barinez:
Small changes for card
https://trello.com/c/W4bbtEFb/6063-snack-bar-does-not-show-full-error
PR - just in case, maybe changes will affect something. MWidth style will also applied if style width exceeds real text width. It seems in current version it works the same.


Barinez:
new branch will be created, material_snackbar.flow was refactored

PR #320, opened 2019-08-09
Fixed 'quit' native for cgi

Coolrunner86:
This fix is intended to make code like
iter(...., \ -> doQuitOnSomeCondition);
quit(0);
work properly. Now such code always returns 0.


Coolrunner86:
Also, can CGI app contain something that needs QCoreApplication's event loop? If no, QApplication::quit() may be omitted, since event loop starts after pending quit check.

PR #324, opened 2019-08-13
Wigi: improve support for RTL

ykuzmenko73:
**In the branch**:
image

image

image

**in the master**:
image
image

image


ykuzmenko73:
[Wrong text alignment in expanders in Arabic](https://trello.com/c/o386YsdH/6016-wrong-text-alignment-in-expanders-in-arabic)

lib/tropic/tropic2form.flow
```diff
@@ -309,10 +309,14 @@ tropic2Acc(b : Tropic, parent : TParentInfo, sheet : Stylesheet, metricsOnly : b
 		TDynamicGroup2(stackChanges, currentStack, combiner) : {
 			fgroupChanges : DynamicBehaviour<[FGroupChange]> = make([]);
 			tropicItems : DynamicBehaviour<[TDynamicGroupItem]> = make([]);
-			verticalCombiner = if (combiner == TLinesCombiner()) FAddition() else FMax();
-			rtl = combiner == TColsACombiner() && getDefaultRtl();
+			verticalCombiner = if (combiner == TLinesCombiner() || combiner == TLinesACombiner()) FAddition() else FMax();
+			rtl = getDefaultRtl();
```
brawaga: I do not remember exactly what for, but surely this condition (rtl) was added for a reason. Did you ensure all combiners work correctly?

brawaga:
imageExcellent.

lib/form/gui.flow
```diff
@@ -490,6 +492,42 @@ Lines(rows) {
...
+			Constructor(
+				Translate(
+					diffWidth,
+					make(0.0),
+					Inspect([Width(widths[i])], l)
+				),
+				\ -> {
+					uns = connectSelect2Distinctu(maxWidthB.first, widths[i], diffWidth, \maxWidth, width -> maxWidth - width);
+					\ -> {
+						if (i == 0) maxWidthB.second();
```
yegordovganich: only if i == 0?

ykuzmenko73: maxWidthB is "global" behavior, we should destroy it only once.
there is another option of course, I could pass it to the Lines level and wrap resulting Lines with another Constructor.

ykuzmenko73: rewritten to be less obstructive

artxp:
@ykuzmenko73 did you test drop down. I remember initial fix affected it behaviour.

PR #325, opened 2019-08-14
Do not trigger click to collapse combobox if the mouse was moved duri…

fsodegov:
@alstrup @anatolmykolenko 
guys, please review. This is to fix the bug when the dropdown list of TCombobox has scroll and was collapsing when we try to move it. I have introduced TDeltaGhosted style. The idea is taken from TRawButton.


anatolmykolenko:
@fsodegov  - looks ok to me. Also i think this TCombobox used only by CCL though

PR #331, opened 2019-08-16
Introduce MSetRTL style.

NikitaFil:
Makes rtl layout for the specific part of content.


lib/material/tests/test_mparagraph_aligns.flow
```diff
@@ -0,0 +1,22 @@
+import material/material;
+import material/material_manager;
+import material/material_dialog;
+
+main() {
+	setLang("ar");
+    manager = makeMaterialManager([]);
+    text = "Science, my boy, is made up of mistakes, but they are mistakes which it is useful to make, because they lead little by little to the truth.";
```
ykuzmenko73: Could you check how "\n" symbol inside the text will be honored?

lib/tropic/tropic_gui.flow
```diff
@@ -58,16 +58,30 @@ export {
 	TGrid(c : [[Tropic]]) -> Tropic;
 
 	// Arabic(RTL) versions
+	TColsRTL(t : [Tropic], isRTL : bool) -> Tropic;
```
ykuzmenko73: We don't have M*RTL functions, just M*A, maybe T*RTL should be internal only? AFAIU, the are not intended  to be used outside tropic's implementation

lib/material/internal/material_grid.flow
```diff
@@ -34,6 +34,7 @@ MFlexibleGrid2T(manager : MaterialManager, parent : MFocusGroup, m : MFlexibleGr
 		|> (\st -> arrayPush(st, TEnabled(en)))
 		|> (\st -> arrayPush(st, TLockWidth(lockWidth)))
 		|> (\st -> arrayPush(st, TCropView(cropView)))
+		|> (\st -> arrayPush(st, TSetRTL(par.rtl)))
```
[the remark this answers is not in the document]
sergiy-stryzhenko: Or we could `concat(st, [TEnabled(en), TLockWidth(lockWidth), etc.])`, couldn't we?

lib/tropic/tropic_gui.flow
NikitaFil: T*RTL are used for Material implementation, so they should be exported here. However, I can add a comment about T*A preferability.

lib/material/internal/material_grid.flow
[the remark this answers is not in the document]
NikitaFil: For sure, concat / concatA is preferable here.

lib/material/tests/test_mparagraph_aligns.flow
NikitaFil: Tested with \n, works fine. 
image

lib/tropic/tropic_gui.flow
brawaga: All this renaming is bad, hence assumes these structs/functions work only with RTL, whilst they implemented to work for both LTR and RTL cases. Must be turned back.

PR #332, opened 2019-08-16
Fixed loss of Japanese charactes in the end of lines which was introduced by hyphenation fixes

artxp:
https://trello.com/c/9lkrdGZO/622-some-japanese-characters-missing-in-paragraphs

### DIFF
--- a/lib/form/paragraph.flow
+++ b/lib/form/paragraph.flow
@@ -681,9 +681,12 @@ breakTextFragment(
 						PositionAllowBreakAfter() : {
 							toInspectingElementIfNecessary(indexing,
 								indexing + keyPosRelative,
-								makeSpace(getCharAt(text, keyPos),
-									style
-								)
+								// in case of CJKLang we don't show the break symbol
+								if (isCJKLang()) {
+									Space(Empty())
+								} else {
+									makeSpace(getCharAt(text, keyPos), style)
+								}
 							);
 						}
 						PositionAllowSoftBreakAfter() : {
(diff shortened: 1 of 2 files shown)


artxp:
paired with https://github.com/area9innovation/innovation/pull/933

artxp:
I broke Japanese functionality when fixed Allowing breaks after some symbols in English.

According to https://github.com/area9innovation/flow9/blob/master/lib/text/localization.flow#L135 we have to differentiate approaches for breaking symbols whether it's CJKLang() or not. So for Japanese (wich is CJKLang) I restored the previous approach.

lib/form/paragraph.flow
```diff
@@ -681,9 +681,12 @@ breakTextFragment(
 						PositionAllowBreakAfter() : {
 							toInspectingElementIfNecessary(indexing,
 								indexing + keyPosRelative,
-								makeSpace(getCharAt(text, keyPos),
-									style
-								)
+								// in case of CJKLang we don't show the break symbol
```
ykuzmenko73: Extract the knowledge about the language's specifics from paragraph.flow into separate file related to languages and use it's functions here.
The function could for example returns special data type with all required field to be used here

artxp: @ykuzmenko73 addressed

artxp:
[Some Japanese characters missing in paragraphs](https://trello.com/c/9lkrdGZO/622-some-japanese-characters-missing-in-paragraphs)

PR #333, opened 2019-08-19
Wigi editor performance tuning in JS

ykuzmenko73:
Try to tune wigi editor in JS mode
deferred option is disabled under NEW flag only. other settings are permanent

1) Hidden Text input in wigi editor should not induce animate
2) Animate should not induce recursive stage's invalidation
3) Do not use deferred in text input pipeline - it lead to several animate calls
4) Do not call html2text for the text in the editor

image

image


ykuzmenko73:
[Performance issues for a customer - Safari on Mac](https://trello.com/c/W6VTjpNh/5934-performance-issues-for-a-customer-safari-on-mac)

ykuzmenko73:
faced performance drop in Chrome in large paragraph case, investigation in progress

ykuzmenko73:
the performance drop happens on abnormal large paragraph only.

PR #335, opened 2019-08-20
Get rid of BytesOutput

yegordovganich:
test:

> 	timeit(\ -> {
		s = "Hello World";
		generate(0, 100000, \__ -> {			
			string2utf8(s) |> ignore			
		}) |> ignore
	}) |> println;

shows 98 ms instead of 189 ms in average

### DIFF
--- a/platforms/js/Native.hx
+++ b/platforms/js/Native.hx
@@ -1690,11 +1690,9 @@ class Native {
 	}
 
 	public static function md5(content: String) : String {
-		var b = new StringBuf();
-		var c = string2utf8(content);
-		for (i in c)
-			b.addChar(i);
-		return Md5.encode(b.toString());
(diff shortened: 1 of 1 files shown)


alstrup:
LGTM

tupieurods:
The main problem remains: although we know final size of the array - we don't provide JS a hint about it in generated code. This can be fixed by using Vector<Int>. But we must return an array. This would lead to a.slice(0) code in JS. Which is again an array copy.

I tried to minimize memory allocations in Native.md5. While it performs better in synthetic tests(no more hang because of possible OOM in chrome) - in real app it still leads to the same error(but memory usage is better too). Inside haxe sources: https://github.com/HaxeFoundation/haxe/blob/3.4.7/std/haxe/crypto/Md5.hx#L112
Because it uses Array and push. And this can't be fixed without changing haxe sources.

PR #339, opened 2019-08-21
Publish Progressive Web App from Content Package: dynamic PWA manifes…

cs1924:
We need to make special PWA for some customers.
To implement this feature we can make manifest depend on skin passed as url flag.
Can we do that?
https://trello.com/c/2WyKOoOA/5159-publish-progressive-web-app-from-content-package


cs1924:
This also means that we can't leave iOS icons statics. There is a library for that
https://developers.google.com/web/updates/2018/07/pwacompat
We need to make sure it works properly.

cs1924:
@vereschagai, @plartem, agree to merge?

PR #340, opened 2019-08-22
Haxe js md5 port

tupieurods:
Removed haxe.io.Bytes.ofString because it was using Array without predefined size.
Encoding conversion code in JsMd5 is based on this part https://github.com/HaxeFoundation/haxe/blob/3.4.7/std/haxe/io/Bytes.hx#L523-L547
Which is haxe.io.Bytes.ofString.
And removed blocks array allocation from MD5 implementation. We don't need more than 64 bytes at a time for a single round - so no need to precompute all of them.
I didn't benchmark old and new implementation yet in terms of speed. But version from this pr doesn't produce extra memory overhead. And this was our original issue.

### DIFF
--- a/platforms/js/JsMd5.hx
+++ b/platforms/js/JsMd5.hx
@@ -1,0 +1,291 @@
+/*
+ * Copyright (C)2005-2017 Haxe Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


platforms/js/JsMd5.hx
```diff
@@ -0,0 +1,302 @@
...
+			bytesInStorage = 1;
+			bytesStorage = 0x80 | (c & 63);
+		} else if (c <= 0xFFFF) {
+			result = 0xE0 | (c >> 12);
+			bytesInStorage = 2;
+			bytesStorage = (0x80 | ((c >> 6) & 63)) | ((0x80 | (c & 63)) << 8);
+		} else {
+			result = 0xF0 | (c >> 18);
+			bytesInStorage = 3;
+			bytesStorage = (0x80 | ((c >> 12) & 63))
```
tupieurods: we can replace 3 ```0x80 |``` with a single ```0x00808080 |```. But i decided not to introduce it right now - to make PR review easier.

alstrup:
LGTM.

tupieurods:
md5 from string executed 100000 times:
String with 25988 English letters in it. master: ~135s. branch: ~91s.
String with 134 English letters. master: ~1065ms. branch: ~780ms.
So we shouldn't see performance drop.

PR #342, opened 2019-08-23
Open last pebble in js;

bvaschenko:
https://trello.com/c/h5rrDNsI/964-restore-the-last-pebble

### DIFF
--- a/lib/pebbles/pebble_controller.flow
+++ b/lib/pebbles/pebble_controller.flow
@@ -49,9 +49,12 @@ export {
 	restoreLastPebble(controller : PebbleController, pebble : Pebble) -> void;
 	// Just refresh the view, i.e. call dispatcher once again. No URL or history changes 
 	refreshViewForCurrentPebble : (controller : PebbleController) -> void;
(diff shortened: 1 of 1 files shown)


lib/pebbles/pebble_controller.flow
```diff
@@ -49,6 +49,9 @@ export {
 	restoreLastPebble(controller : PebbleController, pebble : Pebble) -> void;
 	// Just refresh the view, i.e. call dispatcher once again. No URL or history changes 
 	refreshViewForCurrentPebble : (controller : PebbleController) -> void;
+	// we will find last visited pebble by this key
+	// in case of js and onJS is true - last visited pebble will be open at start without prompt
+	setLastPebbleKey(key : string, onJS : bool) -> void;
```
dmitrybure: Explain how to use this function in the app. Also mentioned that it will be saved by default in CPP case, and `onJS` flag is a way to control it in JS case. Also I would rename "onJS" to "saveLastPebbleInJsCase" or something like this

PR #346, opened 2019-08-28
JS. FileDrop. Prevent native droparea from intercepting clicks.

vol-kharchenko:


vol-kharchenko:
Regarding case https://trello.com/c/Lk9U3Z5a/6383-communication-recipients-cant-be-selected-in-composition-dialog

PR #348, opened 2019-08-31
Implement serializing KeyValues for use in URLs

sergiy-stryzhenko:
Required for Trello card:
https://trello.com/c/3mBgptzZ/6437-invalid-link-url-in-invitation-emails

Add functions that convert a list of `KeyValue`s to a string in format
`application/x-www-form-urlencoded`


sergiy-stryzhenko:
The file where I've added these functions may be wrong, so can anyone come up with a better place?

PR #351, opened 2019-09-05
JS: introduce cloneString fn and tune parseJson / deserializeRestOfStringTail

ykuzmenko73:
In Chrome, when we substring str then we get not new copy but sliced string
if we do many nested slices then we may got extremely long (1000+) slicing chain with size much more that total size of substrings.
it is especially true if we deserialize some recursive structure like Json
this fix breaks slicing on the Object and String level for Json.

checked with string to Json convertion with string size 42 Mb
in master the real size of Json is about 300 Mb, in the branch i is only 65 Mb
the parsing time reduced almost twice.

in the branch:
image

in the master:
image

loading http://localhost/rhapsode/flowjs.html?name=curator&new=1&server=qa#modules:

master: 13.5 sec, 392 Mb after GC

branch: 10 sec, 185 Mb after GC

### DIFF
--- a/lib/formats/json/json.flow
+++ b/lib/formats/json/json.flow
@@ -279,7 +279,7 @@ parseId(s : string, l : int, start : int) -> Pair<string, int> {
 	end = countUntil(start, l, \i -> {
 		getCharCodeAt(s, i) == 58; // ":"
 	});
-	Pair(trim(substring(s, start, end - start)), end);
+	Pair(cloneString(trim(substring(s, start, end - start))), end);
 }
 
 getJsonStringField(json, field, def) {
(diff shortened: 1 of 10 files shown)


ykuzmenko73:
QtByteRunner rebuilt for linux, needed to be rebuilt for windows too

alstrup:
Use fallback native implementation, instead of dummy implementation in all targets but JS/Haxe.

ykuzmenko73:
fallback native implementation is used instead of CPP and others

alstrup:
LGTM.

ykuzmenko73:
[Investigate performance for curator based on prod data](https://trello.com/c/mC0Mhb5G/629-investigate-performance-for-curator-based-on-prod-data)

PR #354, opened 2019-09-09
JS Native json

VsevolodZakharov:
This is 2 times faster with json/test/speed_test.flow
@alstrup , @ykuzmenko73 , please take a look

### DIFF
--- a/lib/formats/json/json.flow
+++ b/lib/formats/json/json.flow
@@ -5,8 +5,8 @@ import text/deserialize_string;
 
 
 export {
-	parseJson(json : string) -> Json;
-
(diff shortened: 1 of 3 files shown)


alstrup:
How does it behave when you give it mal-formed JSON? I know our parseJson does not have any particular error recovery strategy, but worth just checking what happens in these cases.

VsevolodZakharov:
It returns JsonDouble(0) when exception. Most cases doParseJson() does the same - for example for empty input string,

alstrup:
OK, in that case, good with me.

ykuzmenko73:
Checked the branch, it seems it 3-5 times faster that flow-base Json and does not make lot of work for GC

PR #355, opened 2019-09-10
Flowc imports check optimization

dmitry-vlasov:


lib/behaviour.flow
```diff
@@ -63,6 +63,14 @@ export {
 	// Retrieves the leaks
 	getSuspectCallStacks(categoryTag : ?) -> [string];
```
alstrup: We should add a helper function instead of exporting this. We do not want more to inspect it, so better to add a function.

dmitry-vlasov: At first I want to add a getter function, but then I found, that other field of DebuggedSubscriber is also use. So if already 2 fields are used, I thought, maybe it is a sign that this stuct should bu exported?

PR #359, opened 2019-09-11
Add utf8Decode to support Arabic language

yegordovganich:
https://trello.com/c/BCWznbXV/6441-module-filter-is-not-working-on-educator

Added utf8Decode which is based on urlDecode0 but just getCharCodeAt instead of extractPercentHex as Asger suggested, works perfectly.

`
import runtime;
import formats/base64;
import formats/html/html;

main() {
	arabic = "أساسيات مراكز الاتصال";
	// println(arabic);
	expanded = expandUtf8(arabic);
	encoded = encodeBase64(expanded);
	decoded = decodeBase64(encoded);
	result = utf8Decode(decoded);
	println(arabic == result);

	quit(0);
}
`

Prints "true" in the console.

### DIFF
--- a/lib/formats/html/html.flow
+++ b/lib/formats/html/html.flow
@@ -555,7 +555,7 @@ utf8DecodeBytes(s : string, extractor : (string) -> Pair<int, string>) -> Pair<i
 		} else if (hexcode <= 0xF7) { // 4 bytes
 			secondByte = extractor(pair.second);
 			thirdByte = extractor(secondByte.second);
-			fourthByte = extractor(secondByte.second);
+			fourthByte = extractor(thirdByte.second);
 			if (secondByte.first == -1 || thirdByte.first == -1 || fourthByte.first == -1) {
 				// Crap. We need 4 bytes
 				pair
--- a/lib/formats/html/html.flow
+++ b/lib/formats/html/html.flow
@@ -30,6 +30,10 @@ export {
 	// '+' decodes to ' '; '%2B' decodes to '+'
 	urlDecode2(s : string) -> string;
 
+	// Decodes UTF-8 string
+	utf8Decode(s : string) -> string;
+
+
 	// presents string characters as corresponding htmlentity codes
 	encode2htmlentities (s : string) -> string;
 	// The tree contains characters/symbols with their corresponding htmlentity codes.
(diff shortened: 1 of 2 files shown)


alstrup:
I would prefer just to copy the entire thing to a new text/utf8.flow file and not change urlDecode.

lib/formats/html/html.flow
```diff
@@ -490,67 +494,91 @@ urlEncode(s){
...
-					Pair((hexcode - 224) * 4096
-						+ (secondByte.first - 128) * 64
-						+ (thirdByte.first - 128),
-						thirdByte.second
-					)
-				}
-			} else if (hexcode <= 0xF7) { // 4 bytes
-				secondByte = extractPercentHex(percentPair.second);
-				thirdByte = extractPercentHex(secondByte.second);
-				fourthByte = extractPercentHex(secondByte.second);
```
invisicat: btw shouldn't it be fourthByte = extractPercentHex(thirdByte.second); ? this implies a change in derived code too

invisicat:
mine's here #360

lib/formats/html/html.flow
yegordovganich: sure, this is a bug at initial code, fixed, thx

yegordovganich:
@alstrup what is the reason of copy/pasting code? In fact, I didn't touch urlDecode but just refactored it to reuse common part.
Let's move this utf8-related stuff to a separate module and import it at html.flow, OK?

alstrup:
For this code, having a simple code is much easier to read and understand, compared to the "generalized" version which is hard to understand. So I think the copy is warranted when it is this little code that is unlikely to need to change.

yegordovganich:
Let me not agree with you. I see what you mean about increasing complexity but we always sacrifice simplicity against clean code reducing potential bugs. I don't think I can convince you but I've just seen the obvious bug Cyril found at initial code and hopefully he fixed it in both versions. But when he copy-pasted code and modified it he added a new bug https://github.com/area9innovation/flow9/pull/360/files#r323566984 and I don't blame him, it always happens when you copy-pasting, that's why I hate it. Who knows how many bugs more were added to copy-pasted code which were not noticeable at the first glance.

yegordovganich:
I don't want to waste your time here, but that's my opinion. You decide:)

alstrup:
I claim he found the bug because his code was simpler.

PR #360, opened 2019-09-12
utf8 operations in a separate file

invisicat:
also a potential fix in html

### DIFF
--- a/lib/formats/base64.flow
+++ b/lib/formats/base64.flow
@@ -9,9 +9,6 @@ export {
 
 	// Convert binary (using lower 8 bits of each character) to base-64 encoded string
 	encodeBase64(b : string) -> string;
-
-	// Expand string to UTF8
-	expandUtf8(s : string) -> string;
 }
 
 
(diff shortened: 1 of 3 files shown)


alstrup:
LGTM. I prefer this to Yegor's solution. If there is a bug in the urlDecode, then fix that as well.

lib/formats/base64.flow
```diff
@@ -9,9 +9,6 @@ export {
 
 	// Convert binary (using lower 8 bits of each character) to base-64 encoded string
```
alstrup: Do add a comment explaining that if you need this to work with Unicode, then use expandUtf8 from utf8.flow to help the next person

lib/text/utf8.flow
```diff
@@ -0,0 +1,69 @@
...
+				thirdByte = getCharCodeAt(s, at + 2);
+				if (secondByte == -1 || thirdByte == -1) {
+					// OK, it is some crap, because we expect a second and third percent here
+					Pair(-1, 0);
+				} else {
+					Pair((hexcode - 224) * 4096 + (secondByte - 128) * 64 + (thirdByte - 128), 3)
+				}
+			} else if (hexcode <= 0xF7) { // 4 bytes
+				secondByte = getCharCodeAt(s, at + 1);
+				thirdByte = getCharCodeAt(s, at + 1);
```
yegordovganich: looks like here is a new bug, + 2 ?

PR #363, opened 2019-09-16
Flowc jar target

dmitry-vlasov:


alstrup:
LGTM, but maybe we should include a hint that you should have JAVA_HOME in your path if there is an error.

PR #366, opened 2019-09-17
Js peepholes

VsevolodZakharov:
@alstrup , @ykuzmenko73 , please take a look.
initStruct: I remember performance of struct creation was measured for different cases. But for now to me returning object literal is faster everywhere. (In the Edge and Mac Safari inlined obj literal is much more faster somehow).  The tests I used: http://jsben.ch/lVsaS. Flow code which works with struct is ~5% faster with that in Chrome and others (Twice faster in FF) Also JS is ~2% shorter.


ykuzmenko73:
Looks good to me

VsevolodZakharov:
@alstrup , can we also require ES6? It is old enough now. https://caniuse.com/#search=ecmascript%206

alstrup:
LGTM, assuming it was tested. We might still have IE11 users, so ES6 in general is maybe still too early. But some ES6 features might be relevant as long as IE11 works.

PR #385, opened 2019-10-04
Wigi table: allow TightWidth and non-start align

ykuzmenko73:
https://github.com/area9innovation/innovation/pull/1076


ykuzmenko73:
[Fit width does not support right justified](https://trello.com/c/CjjPPxAH/597-fit-width-does-not-support-right-justified)

PR #387, opened 2019-10-07
Flowc proxy removed

dmitry-vlasov:


alstrup:
If it works, great. Maybe ask those that reported problems to try it?

PR #390, opened 2019-10-08
Flowc ping server

dmitry-vlasov:


alstrup:
Looks reasonble. Next step is to use ping in parallel to detect dead servers. If it does not respond without normal time, we should probably just assume the server is dead, and attempt to kill and restart it.

tools/flowc/flowc_usage.flow
```diff
@@ -151,6 +151,12 @@ printUsage(config) {
 	prih(help,"      output-dir=<dirname>      Use a specified directory to store built targets");
 	prih(help,"      working-dir=<dirname>     The directory, relative to which files are resolved");
 	prih(help,"      compile-expression=<exp>  Compile the expression and return corresponding serialized internal representation");
+	prih(help,"      ping=1                    Ping a server");
+	prih(help,"      ping-timeout=<ms>         Timeout for a ping request, default is 5000");
+	prih(help,"      request-timeout=<ms>      Timeout for any other request, default is 0 (infinity)");
+	prih(help,"      watchdog-mode=1           Start a watchdog, which starts flowc server, keeps it unning and prevents hanging / overloading");
```
alstrup: "running" with r

dmitry-vlasov: fixed

PR #394, opened 2019-10-10
Dynamic typing in wigi

chekaninmaxim:


chekaninmaxim:
[Wigi editor: we should not recreate all UI elements in paragraph on every change in text](https://trello.com/c/n7xCNmis/625-wigi-editor-we-should-not-recreate-all-ui-elements-in-paragraph-on-every-change-in-text)

PR #396, opened 2019-10-15
Escape pebble values

iv-menshenin:
https://trello.com/c/oyem9jV0/7072-communicator-screen-doesnt-render-correctly-as-you-click-through-the-messages


iv-menshenin:
@bvaschenko Oops, sorry. I reverted changes for the binaries

dmitrybure:
@bvaschenko don't you think having other special characters might cause problems with browsers? but it depends on how functions like `setUrlHash` are implemented

bvaschenko:
> @bvaschenko don't you think having other special characters might cause problems with browsers? but it depends on how functions like `setUrlHash` are implemented

https://tools.ietf.org/html/rfc3986#section-3.5
It shouldn't cause any problem as URL-encoding is necessary for passing parameters to the server, but URL hash is used only at the client-side.

dmitrybure:
OK, then let's try `escapePebbleParameterValue`

bvaschenko:
https://github.com/area9innovation/flow9/commit/bb2144589cbbeb8854159f655d478c1722ae6a22

PR #398, opened 2019-10-21
Use concat instead of arrayPush

anastasiakucherova:
Better to apply concat here because arrayPush is not optimal in case of using convertTreeToArray etc. in fold/recursion

### DIFF
--- a/lib/ds/tree.flow
+++ b/lib/ds/tree.flow
@@ -565,13 +565,13 @@ sizeTree(t) {
 
 treePushToArrayValue(tree, key, value) {
 	c = lookupTreeDef(tree, key, []);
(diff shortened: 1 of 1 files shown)


alstrup:
I don't get this. Why would concat be better?

anastasiakucherova:
I was confused by the comment https://github.com/area9innovation/flow9/blob/master/lib/ds/array.flow#L46
Some time ago I faced with problem when arrayPush was inside fold/recursion and there was really big memory consumption.
But now it seems arrayPush and concat behave the same way - I did a few tests for js and cpp targets, everything's OK.
Sorry for disturbing :(

PR #401, opened 2019-10-23
checking of the SW version, checking of the SW updates

MalkoVl:
Added SW versions check on SW registration.
Added listener for SW updates.
More info about SW in the console.
image
image
image


alstrup:
Where do we reference sw.js? It sounds like those places should have a "sw.js?v2" URL or similar instead, so we avoid this entire thing and can force a reload.

MalkoVl:
It's hardcoded on native: https://github.com/area9innovation/flow9/blob/master/platforms/js/ProgressiveWebTools.hx#L14
Not sure, but if we will register new SW (with new path) - it will be another SW and previouse cache will be lostю

MalkoVl:
Also, Ivan has improved SW code by adding SW restart if new version detected (without app restart). For me, SW was restarted in about 4-5 minutes after application was loaded.

MalkoVl:
@alstrup does we need it (mostly for investigating sw issues on remote servers) or Ivan's changes is enough?

alstrup:
The simpler, the better. I can not evaluate if these changes help or not. If Ivan's changes are sufficient, then we do not need this.

MalkoVl:
We decided to have it.

alstrup:
Let us change the message to

"A component is updated. Please reload the application to apply it."

PR #402, opened 2019-10-23
Do not add jusify offset after first word in listed paragraph

chekaninmaxim:


chekaninmaxim:
[Bullets display is off on Iphone 5](https://trello.com/c/p3jYFxBe/7197-bullets-display-is-off-on-iphone-5)

PR #403, opened 2019-10-24
Flowc fix module cache

dmitry-vlasov:


alstrup:
Hard to review, but seems correct in principle. How was this tested?

dmitry-vlasov:
I tested it locally, on my machine. The fact that server drops it cache when objc folder is deleted I could see in server console output: it start parsing of all file

PR #406, opened 2019-10-30
Wigify TightWidth and Align - rework using WidthManager

ykuzmenko73:
Wigify worked wrong in case of TightWidth and non-start alignment because of a bug - renderParagraph used Available Width inspector to spread words in the lines according to Align option, but this value is global Available Width. In case of TightWidth it should be calculated on the parent story level.


ykuzmenko73:
[Fit width does not support right justified](https://trello.com/c/CjjPPxAH/597-fit-width-does-not-support-right-justified)

lib/form/paragraph3.flow
```diff
@@ -893,3 +915,20 @@ setWordText(w: ParaWord, t: string) -> ParaWord {
 setWordForm(word: ParaWord, form: GhostForm) -> ParaWord {
 	ParaWord(word.word, word.metrics, word.ghosted, form, word.interactivityIdM);
 };
+
+makeWigiStoryWidthManager(paragraphs : int, tightWidth : bool) -> WigiStoryWidthManager {
+	paragraphWidths = generate(0, paragraphs, \__ -> make(0.0));
+	if (tightWidth) {
+		storyWidth = arrayMaxu(paragraphWidths);
```
chekaninmaxim: Looks like you missed to unsubscriber of arrayMaxu, or I missunderstand something.

ykuzmenko73: I didn't call unsubsriber, it is true. There is no good and easy to use place to call it.
We create WigiStoryWidthManager in the doWigifyStory but the real form is created later and we can't call disposer in the usual place like form's Constructor's destructor because it would be called when the form is hided and then after shows the subscription will be lost.

but I think that this will not introduce memory leak because GC will collect all this elements at at once when the WigiStoryWidthManager is collected.
maybe I'm wrong. @yegordovganich , what do you think, will it produce memory leak or not?

lib/form/paragraph3.flow
```diff
@@ -163,6 +182,9 @@ renderParagraph(p : [ParagraphElement], s : [ParagraphStyle], rtl : bool) -> For
 				// It's the safest way to simmulate single line, because reflowParaWords2 does a lot of useful staff.
 				w = if (isSingleLine) 100000.0 else max(0.0, wi);
 				newLines = reflowParaWords(words, w, indent);
+
+				if (hasWidthManager) nextDistinct(widthManager.width, fold(map(newLines, \l -> fold(l.words, 0.0, \acc, w1 -> acc + geParaWordWidth(w1))), 0.0, \acc, width -> if (width > acc) width else acc));
```
artxp: too long line

ykuzmenko73: addressed, reworked to use already calculated width

lib/form/paragraph3.flow
artxp: I don't think it's good idea to leave it w/o unsubscriber call. The unsubscriber is not necessary to be called by constructor, but manually.

ykuzmenko73: @alstrup , I need your advice about if the unsubscribe call really required in this case.
If I'm not mistaken, it is redundant here because subscription happens inside function makeWigiStoryWidthManager and arrayMaxu is used on behaviors created there too, so when the object WigiStoryWidthManager will be available for GC it collects it along with subscription. The WigiStoryWidthManager will be ready for GC when the resulted MWigify material will be removed from the screen.

ykuzmenko73: fixed

PR #408, opened 2019-11-02
Flowc bug with array inds

dmitry-vlasov:


tools/flowc/typechecker/combine_types.flow
```diff
@@ -292,23 +292,27 @@ combinePositiveAndNegative(env : FcTypeEnv, positive : FcType, negative : FcType
...
-								mapi(pargs, \i, parg -> {
-									FcStructArg(
-										parg.name,
-										combinePositiveAndNegative(env, parg.type, sargs[i].type, onError)[0],
-										parg.ismutable
-									)
-								}),
-								si
-							)]
+							pargs_combined = mapi(pargs, \i, parg -> {
```
alstrup: filtermapi, and then check the length instead of first map, then forall, then filtermap.

PR #409, opened 2019-11-06
pebbles - allow custom hierarchy item name and ability to hide chips

aatkachenko:
This is to reach 2 goals:
1) include in pebble hierarchy items with parameter names already in use which we should not change because they are used in notification emails being sent.
2) keep hierarchy parameters in a view sequence like this
v1 -> v3 -> v2 
where v1, v2 use chips and hierarchy filtering 
but v3 uses breadcrumbs and do not need the hierarchy filtering

### DIFF
--- a/lib/pebbles/pebble_hierarchy.flow
+++ b/lib/pebbles/pebble_hierarchy.flow
@@ -26,7 +26,7 @@ export {
 	);
 
 	makePebbledHierarchyItem(name : string, id2label : (int) -> string, children : [PebbledHierarchyItem]) -> PebbledHierarchyItem;
-	makePebbledHierarchyItem2(name : string, customHierarchyParameterName : string, id2label : (int) -> string, children : [PebbledHierarchyItem]) -> PebbledHierarchyItem;
+	makePebbledHierarchyItem2(name : string, customParameter : string, id2label : (int) -> string, children : [PebbledHierarchyItem]) -> PebbledHierarchyItem;
 
 	// Build a material corresponding to the current values of pebbled hierarchy parameters.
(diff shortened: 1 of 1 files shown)


lib/pebbles/pebble_hierarchy.flow
```diff
@@ -26,6 +26,7 @@ export {
 	);
 
 	makePebbledHierarchyItem(name : string, id2label : (int) -> string, children : [PebbledHierarchyItem]) -> PebbledHierarchyItem;
+	makePebbledHierarchyItem2(name : string, customHierarchyParameterName : string, id2label : (int) -> string, children : [PebbledHierarchyItem]) -> PebbledHierarchyItem;
```
dmitrybure: customHierarchyParameterName -> customParameter

lib/pebbles/pebble_hierarchy.flow
```diff
@@ -36,9 +37,10 @@ export {
 		style : [PebbledHierarchyBodyStyle]
 	) -> Material;
 
-	PebbledHierarchyBodyStyle ::= PebbledHierarchyBodyAdditionalChips, PebbledHierarchyBodyErrorView;
+	PebbledHierarchyBodyStyle ::= PebbledHierarchyBodyAdditionalChips, PebbledHierarchyBodyErrorView, PebbledHierarchyBodyHideChips;
 		PebbledHierarchyBodyAdditionalChips(chips : Transform<[Material]>);
 		PebbledHierarchyBodyErrorView(fn : (itemName : string, itemId : string) -> Material);
+		PebbledHierarchyBodyHideChips();
```
dmitrybure: add a comment explaining the purpose of this style

PR #414, opened 2019-11-11
MTree. Added MTreeLazyLoad.

NikitaFil:


danichmur:
@ NikitaFil it works very fast, thanks! There are two things:
1) can we make the button in the same style as other children?
! [Image] (https://user-images.githubusercontent.com/26232105/68647576-47893e80-052f-11ea-8428-57f2f618c349.png)
2) can we reset the state when the parent was reduced?

NikitaFil:
@danichmur 
1. You can create custom panel by buildPanelFn parameter.
2. It`s a good suggestion. I`ll implement it.

PR #418, opened 2019-11-18
binarySearch fix

Undiy:
Binary search didn't handle the case with insert second (in front or at the end of array).
Here's a gist with a test case: https://gist.github.com/Undiy/f0f1c1e4e6dc34e694d685b0e0b71e1b


alstrup:
Lgtm

PR #420, opened 2019-11-20
Fixed extra line break not to be added in Simulate between bullet and text for narrow columns

artxp:
https://trello.com/c/Lsdc8Xpt/7588-extra-line-break-is-added-in-simulate-between-bullet-and-text-for-narrow-columns


artxp:
image

PR #422, opened 2019-11-21
attempt to fix HashMap

shkel:
image

image


lib/ds/hashmap.flow
```diff
@@ -102,26 +102,44 @@ makeHashMapOfVolume(hash : (?) -> int, volume : int, loadFactor : double) -> Has
...
 }
 
 containsKeyHashMap(hashMap : HashMap<?, ??>, key : ?) -> bool {
+	index = hashMapIndex(hashMap, key);
+	if (existsIndex(hashMap.data, index)) {
+		containsKeyTree(^(hashMap.data[index]), key)
+	} else {
+		false
+	}
 	containsKeyTree(^(hashMap.data[hashMapIndex(hashMap, key)]), key)
```
alstrup: Should be removed? Consider && instead of if

shkel: sorry. fixed

alstrup:
Good

PR #426, opened 2019-11-22
Java add missing native stubs

dmitry-vlasov:


lib/formats/html/html.flow
```diff
@@ -458,37 +458,42 @@ doUnescapeHtml(acc : string, s : string, numericOnly : bool) -> string {
...
-	|> r("@","%40")
-	|> r("=","%3D")
-	|> r("&","%26")
-	|> r("$","%24")
-	|> r("'","%27")
-	|> r("–","%E2%80%93")
-	|> r("™", "%E2%84%A2")
-	|> r("\n", "%0a")
+	pairs2tree([
+		Pair(char2int("%"), "%25"), // important to be first
```
alstrup: Notice the comment. It says that one has to be first. It can not be correct to use a tree where the order is changed. We have a multi-string replacement function, which uses an array. Why not just use that?

dmitry-vlasov: ok, I'll correct it

PR #428, opened 2019-11-27
pairs2hashMapArray

shkel:


lib/ds/hashmap.flow
```diff
@@ -221,6 +222,16 @@ pairs2hashMapWithLoadFactor(pairs : [Pair<?, ??>], hash : (?) -> int, loadFactor
 	hashMap
 }
 
+pairs2hashMapArray(pairs : [Pair<?, [??]>], hash : (?) -> int) -> HashMap<?, [??]> {
+	len = cast(cast(length(pairs) : int -> double) / defaultHashMapLoadFactor : double -> int) * 2;
```
alstrup: i2d instead of cast

alstrup: and floor the other way

lib/ds/hashmap.flow
```diff
@@ -221,6 +222,16 @@ pairs2hashMapWithLoadFactor(pairs : [Pair<?, ??>], hash : (?) -> int, loadFactor
...
+	hashMap = makeHashMapOfVolume(hash, len, defaultHashMapLoadFactor);
+	iter(pairs, \pair -> {
```
alstrup: Why is this not a fold? Isn't the hashmap functional?

shkel: copy-paste from pairs2hashMap()

lib/ds/hashmap.flow
```diff
@@ -221,6 +223,18 @@ pairs2hashMapWithLoadFactor(pairs : [Pair<?, ??>], hash : (?) -> int, loadFactor
...
 }
 
+pairs2hashMapArray(pairs : [Pair<?, [??]>], hash : (?) -> int) -> HashMap<?, [??]> {
+	len = floor(i2d(length(pairs)) / defaultHashMapLoadFactor) * 2;
+	hashMap = makeHashMapOfVolume(hash, len, defaultHashMapLoadFactor);
+	fold(pairs, hashMap, \newHashMap, pair -> { 
+		setHashMap(
+			newHashMap,
+			pair.first,
+			eitherMap(lookupHashMap(hashMap, pair.first), \arr -> concat(arr, pair.second), pair.second)
```
alstrup: lookupHashMapDef  seems like your friend: Just return the empty array as def, and then you can concat?

shkel: lookupHashMapDef + concat produces less garbage?

alstrup: Probably similar garbage but easier to read

PR #429, opened 2019-11-28
snakeCase - implement cache

ykuzmenko73:
I think it is safe to use cache here because the number of fields is limited and not large, but snakeCase called many times

### DIFF
--- a/lib/string_utils.flow
+++ b/lib/string_utils.flow
@@ -1,4 +1,5 @@
 import string;
+import binarytree;
 
 export {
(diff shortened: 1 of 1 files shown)


ykuzmenko73:
[Investigate performance of learner based on prod data](https://trello.com/c/JHEVQjw6/635-investigate-performance-of-learner-based-on-prod-data)

alstrup:
Does not look like a good idea. This cache is never cleared, so it is too much and will just bloat memory for other users. Why is this a bottleneck? Better to look at the user of this function instead.

PR #433, opened 2019-11-29
fallbacks for concurrent hash map are added to concurrent

dmitry-vlasov:


lib/sys/concurrent.flow
```diff
@@ -129,3 +129,34 @@ concurrentAsyncOne(fine : bool, task : () -> ?, callback : (?) -> void) {
 getThreadId() -> string {
 	""
 }
+
+// Fallbacks for concurrent hash map
+
+initConcurrentHashMap() -> native {
+	cast(ref makeTree() : ref Tree<?, ??> -> flow)
```
alstrup: Use "flow()" instead of cast.

PR #439, opened 2019-12-03
Lingo conditional

Coolrunner86:
Added basis of the preprocessor into Lingo parser. Conditional grammar compilation.


alstrup:
You have to add documentation to doc/lingo.markdown

Coolrunner86:
> You have to add documentation to doc/lingo.markdown

Done

doc/lingo.markdown
```diff
@@ -151,6 +151,31 @@ The grammar
...
+	prod = 'a'* { A() };
+	#elif CONDITION_2
+	prod = 'b'* { B() };
+	#else
+	prod = 'c'* { C() };
+	#endif
+
+Nested conditions are also allowed.
+Definitions can be stated through `definitions` parameter for pegcompiler or by using `compilePegGrammar3` function.
+
```
alstrup: It is not clear whether you can do

prod = something
#ifdef CONDITION
    | more
#else
#endif

Or is it only for complete productions?

Coolrunner86: It's only for complete production. But if there is a need in such functionality, it can be implemented.

alstrup: Then document that limitation, and then OK to merge.

Coolrunner86: In fact, this is rather middle-processor than preprocessor. We could implement stand-alone version of it and use it for other file types.

alstrup:
Yes, that was my original thought, and that seems to be much simpler.

PR #443, opened 2019-12-06
Need to change DynamicBehaviour to Transform in some function arguments for pebble_utils

maxleontev:
We don't need to change the itemsB : DynamicBehaviour inside these functions, but just fgetValue of it.

Also, I need to call these functions using result of 
fselect2(DynamicBehaviour<>, DynamicBehaviour<>, FLift2(...)) as argument, it will looks fine in my code and I would not create unnecessary unsubscribers for it.
But compiler didn't allowed me to do this.

Do you have any objections for this changings?


dimabure:
LGTM

PR #445, opened 2019-12-06
Basic general-purpose preprocessor

Coolrunner86:


Coolrunner86:
Two last commits are from previous PR, since the work was done anyway.

Coolrunner86:
Also I didn't write docs yet.

Coolrunner86:
I wonder if we should move it outside lingo's folder, since it is stand-alone.

alstrup:
Makes sense

PR #448, opened 2019-12-18
inline console.log for correct links

VsevolodZakharov:
@alstrup , @dmitry-vlasov , don't you mind this minor fix to show actual links for messages sources in JS console?


dmitry-vlasov:
On my point of view this change is ok.

PR #449, opened 2019-12-18
Flowc new concurrent

dmitry-vlasov:


alstrup:
Not sure what the goal of this is, but looks reasonable. Would be useful to document what platforms the new natives work on

PR #450, opened 2019-12-18
Some tweaks to tropic paragraph, so it can be used in wigi recursives

chekaninmaxim:


chekaninmaxim:
[Implement wigi recursisves on tropic](https://trello.com/c/SH8Ybtxx/673-implement-wigi-recursisves-on-tropic)

PR #452, opened 2019-12-23
fix wrong metrics in IE

ykuzmenko73:
This PR fixes lost of precision in IE for measuring the text.

image

image

image


ykuzmenko73:
[Wigify(...) in Internet Explorer has larger width than available. Causes wrong scroll in some cases.](https://trello.com/c/8RlnEUeu/669-wigify-in-internet-explorer-has-larger-width-than-available-causes-wrong-scroll-in-some-cases)

ykuzmenko73:
Checked that in real curator this fix breaks everything, looks like font's metrics is not proportional to font size in common case
image
image

ykuzmenko73:
looks like it is possible to fix the issue for IE on the level of workaroundTextMetrics

PR #458, opened 2020-01-10
Flowc cpp2 backend 1

dmitry-vlasov:
flowc may be successfully compiled (bootstrapped) from scratch with itself with the new cpp2 backend.


platforms/qt/main.cpp
```diff
@@ -613,6 +613,18 @@ int main(int argc, char *argv[])
 #ifdef NATIVE_BUILD
         FlowRunner.Init(load_native_program());
         FlowRunner.setUrl(params);
+        QStringList args = app->arguments();
```
alstrup: Add a comment explaining what this does

dmitry-vlasov: Done

PR #474, opened 2020-01-30
Flowc subtyping made a bit less strict

dmitry-vlasov:


tools/flowc/incremental/fc2fi.flow
```diff
@@ -239,7 +239,7 @@ fctype2fi(e : FcTypeEnvLocal, onError : (string, FcInfo2) -> void, t : FcType) -
 		FcTypeVar(tv_, i): {
 			tv = getResolvedFcType(e, t);
 			if (tv == t) {
-				if (!isUrlParameterTrue("gtype")) {
+				if (!isUrlParameterTrue("gtype") && isUrlParameterTrue("strict")) {
 					onError("Could not resolve tyvar " + i2s(tv_), i);
```
alstrup: While this might be OK, I wonder if we get problems in some of the backends due to this. They might complain themselves, of course, so I guess it is OK...

dmitry-vlasov: This is necessary to compile smartbuilder, because there are some unresolved tyvars. Unfortunately it's extremely hard to find out the exact place, where it comes from, because i : FcInfo2 contains only zeros and is generated from FType, which doesn't contain info about origins of entities.

PR #479, opened 2020-02-07
show zoom icons on small screens

shkel:
I am in doubt ... Can I change the current function? (Or create a new one? :( )

https://trello.com/c/MdQ2jQv7/8471-images-always-allow-zoom-although-option-is-disabled#comment-5e3a906e2b470d724e411037

image


alstrup:
Should be fine to change the existing one.

NikitaFil:
Don`t familiar with this part of code, but looks like a proper fix.

PR #483, opened 2020-02-13
Fix promises

yegordovganich:
Warning instead of assert, refactoring and small optimizations
- don't fire until all promises are done (either fulfilled or rejected)
- minor refactoring

### DIFF
--- a/lib/promise.flow
+++ b/lib/promise.flow
@@ -177,31 +177,22 @@ allP0(promises : [Promise<?, ??>], idx : int, acc : List<?>, fulfill : ([?]) -> 
 }
 
 parallelP(promises : [Promise<?, ??>]) -> Promise<[?], ??> {
-	count = ref 0;
-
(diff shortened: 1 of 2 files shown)


yegordovganich:
The problem that in master it crashes when the number of fulfill calls in one promise equals the number of promises - because there is only dull counter.
`promisesCrash = [
  Promise(\f, r -> timer(500, \ -> f(1))),
  Promise(\f, r -> {f(2); f(3)})
`
Timer in promise is an emulation of request to database. So assert crashes the app when f(3) is called. In the branch there is no crash, just waiting for the first promise and fires with [1, 3].

shkel:
image

yegordovganich:
@alstrup I know you are on phone, it's not urgent, but have a look when you have a time, pls

alstrup:
Hard to review, but if tested OK

PR #484, opened 2020-02-14
Set escape html for text field in renderform too

chekaninmaxim:


chekaninmaxim:
[Strings like "&#8747;" are displayed incorrectly in MWEditor on JS target](https://trello.com/c/g7U3q4Aw/689-strings-like-8747-are-displayed-incorrectly-in-mweditor-on-js-target)

PR #491, opened 2020-02-25
Flowc lsp

dmitry-vlasov:


lib/sys/system.flow
```diff
@@ -21,4 +21,13 @@ export {
 
 	// start a process detached from the application process
 	native startDetachedProcess : (command : string, args : [string], currentWorkingDirectory : string) -> bool = Native.startDetachedProcess;
+
+	// Read n bytes from stdin input
+	native readBytes : io (n : int) -> string = Native.readBytes;
```
alstrup: What encoding is used? How are bytes represented in the string? Is it 8-bits in the lower parts, is it interpreted as UTF-8, what is the encoding?

tools/flowc/flowc_usage.flow
```diff
@@ -144,9 +145,10 @@ printUsage(config) {
 	prih(help,"      outfolder=<FOLDER>        Save the typed result of all test cases in the given folder");
 	prih(help,"      verify-types=0            Do not verify the assigned types after typechecking is done");
 	prih(help,"");
-	prih(help,"      server-mode=1             Run flowc in a server mode");
```
alstrup: Is the VS code plugin updated accordingly? Why rename this?

tools/flowc/flowc_usage.flow
```diff
@@ -144,9 +145,10 @@ printUsage(config) {
...
+	prih(help,"      http-server-mode=1        Run flowc as an http server");
 	prih(help,"      server-port=<num>         A port, used in a server mode. Default is '10001'");
 	prih(help,"      server-shutdown=1         Pass this parameter in config in order to stop server");
+	prih(help,"      console-server-mode=1     Run flowc as a console server (uses console for interaction with IDE)");
```
alstrup: Oh, I see. Why not just add "server-mode=http" instead?

tools/flowc/incremental.flow
```diff
@@ -25,7 +25,7 @@ export {
 	mergeIncremental2TypeEnvProgram(env : FcTypeEnvProgram, globEnv : FcTypeEnvGlobal, module : FiModule) -> FcTypeEnvProgram;
 }
 
-incrementalVersion = 6;
+incrementalVersion = 7;
```
alstrup: What is new in the object files?

tools/flowc/incremental/fi_helpers.flow
```diff
@@ -558,3 +569,52 @@ fiTypeIsFlow(t : FiType) {
 		default: false;
 	}
 }
+
+fiModuleDecls(m : FiModule) -> [FiDeclaration] {
+	decls = ref [];
+	iter(m.functions, \f -> refArrayPush(decls, f));
```
alstrup: This has terrible performance and is imperative. Use concatA instead:
  concatA([m.functions, m.globalVars, m.natives])
Shorter, quicker, functional. Same below.

tools/flowc/lsp_types.flow
```diff
@@ -0,0 +1,189 @@
+import formats/json/json;
+import formats/uri2/uri2;
+
+export {
```
alstrup: Add a comment with the LSP link and description.

lib/sys/system.flow
dmitry-vlasov: These functions operate on a byte representation of strings - read n bytes, and compare a byte string stream with a pattern byte-wise. The resulting string in java is formed from a byte array as:
`byte[] bytes = new byte[line.size()]; .... return new String(bytes);`
So it uses a default encoding.
Maybe here I should rather use `return new String(bytes, "UTF-8");` - I'm not sure.

tools/flowc/incremental/fi_helpers.flow
dmitry-vlasov: Fixed.

tools/flowc/lsp_types.flow
dmitry-vlasov: Added.

tools/flowc/flowc_usage.flow
dmitry-vlasov: Yes, flow vscode plugin is updated, a new version is compiled. Users need to re-install it.

tools/flowc/incremental.flow
dmitry-vlasov: I added a new field to FiFileInfo struct:
`mutable source : string, // is loaded by demand. Initially is an empty string.`
because hover feature may demand frequent accesses to the source code (as a string), so it's better to have it cached in memory.

tools/flowc/flowc_usage.flow
dmitry-vlasov: I changed this to:
`server-mode=<http|console> Run flowc as an http server or a console server accordingly to the option value`

lib/sys/system.flow
alstrup: The important thing is to document what it does.

PR #497, opened 2020-02-27
Cartesian product data structure with appropriate functions is added

dmitry-vlasov:


lib/ds/prod.flow
```diff
@@ -0,0 +1,114 @@
+import ds/array;
+
+// Cartesian product of arrays: if c = [a_1, a_2, ... , a_n] then
+// c is interpreted as a_1 x a_2 x ... x a_n
+
+export {
+	// Cartesian product of sets (elements of the array).
+	Prod : (sets : [[?]]);
```
alstrup: Too short a name. Make it longer.

lib/ds/prod.flow
```diff
@@ -0,0 +1,114 @@
...
+
+	isEmptyProd(prod : Prod<?>) -> bool;
+
+	// Map all tuples of product. The result has sizeProd(prod) elements.
+	mapProd(prod : Prod<?>, f : ([?]) -> ??) -> [??];
+
+	// Traverse all tuples of product.
+	iterProd(prod : Prod<?>, f : ([?]) -> void) -> void;
+	foldProd(prod : Prod<?>, init : ??, f : (??, [?]) -> ??) -> ??;
+	fltermapProd(prod : Prod<?>, f : ([?]) -> Maybe<??>) -> [??];
```
alstrup: filter with i

lib/ds/prod.flow
dmitry-vlasov: Renamed to CartProd

lib/ds/prod.flow
dmitry-vlasov: fixed

PR #505, opened 2020-03-11
Flowc java lookup priority

dmitry-vlasov:


alstrup:
Do we have a test case? If so, useful to add. Otherwise, LGTM

PR #506, opened 2020-03-11
Reduce xliff

pzotov:
Task https://trello.com/c/ZxG9DQgR/8513-target-is-not-required-tag-in-xliff-can-we-reduce-xlf-files


lib/text/translationutils.flow
```diff
@@ -211,7 +211,7 @@ extendedTexts2XliffContent(texts : [GeneralContentUnit], productName : string, l
 			text = if (escapedText) xmlEscapeText(t.text) else t.text;
 			"   <trans-unit id=\"" + getId() + "\" " + addAttr(t) + ">\n"
 			"    <source>" +  text + "</source>\n"
-			"    <target>" +  text + "</target></trans-unit>\n"
```
vladimirgorky: I agree that this is a step in right direction but are you sure we don't break anything in old stuff?
I would suggest to add one more parameter or style to extraction functions (and special parameter to build) for this change and try it for example on rhapsode only. And after that apply this to other innovation projects and ccl.

pzotov: I have added the parameter reducedXliff. It is true only for wigiLightTexts2XliffContentWithLang() which is called from Rhapsode.

PR #512, opened 2020-03-20
Feature/395 incorrect behaviour of tcols in rtl mode

brawaga:
This is not an ultimate solution, hence further work should be done on Inspectors. BTW, paragraph rewritten to measure lines first, and then applying alignment when needed width is known; also, alignment switches replaced with simple arithmetics.
TParentInfo has now one more field to know text flow direction in current context and transition of it from MFocusGroup.


lib/form/paragraph3.flow
```diff
@@ -45,6 +45,20 @@ possible, and because rendering words in the same sprite improves the interword
...
 
+// How extra space of paragraph line will be spread:
+// start — added at start of first word;
+// justify — equally distributed between words or added at the end(!) of the only word or last line;
+// end=1-start-justify — what's left will be at end of the line.
+// We might store start and end, but even this variant seems symmetrical,
+// the used one is more useful for calculations.
+GeneralAlignment(
+	start : double,
+	justify : double
```
ykuzmenko73: justify seems to be pure bool, not double.

lib/form/paragraph3.flow
```diff
@@ -45,6 +45,20 @@ possible, and because rendering words in the same sprite improves the interword
...
 
 
+// How extra space of paragraph line will be spread:
+// start — added at start of first word;
+// justify — equally distributed between words or added at the end(!) of the only word or last line;
+// end=1-start-justify — what's left will be at end of the line.
+// We might store start and end, but even this variant seems symmetrical,
+// the used one is more useful for calculations.
+GeneralAlignment(
+	start : double,
```
ykuzmenko73: using double to model 3  way value (0 / 0.5 / 1) looks as old C-style / Form-style, maybe there is better solution?

lib/form/paragraph3.flow
```diff
@@ -143,6 +168,9 @@ renderParagraph(
 	words : [ParaWord] = map(list2array(glued), makeParaWord);
 
 	aw = make(0.0);
+	width = if (widthManager == dummyWigiParagraphWidthManager) make(100000.0) else widthManager.width;
+	next(width, 100000.0);
```
ykuzmenko73: define common variable and use in the code.

ykuzmenko73: why next, not nextDistinct?

lib/form/paragraph3.flow
```diff
@@ -151,16 +179,18 @@ renderParagraph(
 		topLineBaseline, interactiveStyles, rtl, isParagraphMarked
 	);
 
+	maxW = renderParaLinesFn(
```
ykuzmenko73: double rendering is bad. do not do it

lib/form/paragraph3.flow
```diff
@@ -197,20 +230,31 @@ renderParagraph(
...
+				[AvailableWidth(aw)], 
+				if (paragraphBorder.top > 0.0 || paragraphBorder.bottom > 0.0) {
+					Border(0.0, paragraphBorder.top, 0.0, paragraphBorder.bottom, contentGroup);
+				} else {
+					contentGroup;
+				}
+			)
+			// Available2, Size2 — neither works, extra space allocated,
+			// invalid behaviour, visible only when other fillers present in TCols.
+			/*Select(aw, \awv -> {
```
ykuzmenko73: what is it aboud?

lib/form/paragraph3.flow
```diff
@@ -197,20 +230,31 @@ renderParagraph(
...
+			inspector;
+		},
 		\ -> {
 			updateFn = \wi-> {
 				// Ignore negative widths
 				// It's the safest way to simmulate single line, because reflowParaWords2 does a lot of useful staff.
-				w = if (isSingleLine) 100000.0 else max(0.0, wi);
+				w = if (isSingleLine) 100000.0 else max(0.0, min(maxW, wi));
+				nextDistinct(size, WidthHeight(w, 0.0));
+				next(size, WidthHeight(w+1.5, getValue(height)));
```
ykuzmenko73: why? nextDistinct and immediately next?

lib/form/paragraph3.flow
```diff
@@ -197,20 +230,31 @@ renderParagraph(
 		})
 	));
 
+	size = make(WidthHeight(maxW, 0.0));
```
ykuzmenko73: size? where do you use it?

lib/form/paragraph3.flow
```diff
@@ -630,160 +674,155 @@ keepFormTogether(form : Form) -> bool {
...
 	tightWidth : bool,
-	alignment: ParaLineAlignment,
+	alignment: GeneralAlignment,
 	interlineSpacing : double,
 	topLineBaseline : bool,
 	styleTree : Tree<int, [ParaElementInteractiveStyle]>,
 	rtl : bool,
 	isMarked : bool
-) -> Pair<Form, double> {
+) -> Triple<Form, double, double> {
```
ykuzmenko73: Even Pair is not clear. Tripple is really bad. do not use Tripple, make special data type instead

lib/form/paragraph3.flow
```diff
@@ -630,160 +674,155 @@ keepFormTogether(form : Form) -> bool {
...
 }
 
-ParaLineResult(form : Form, width : double, height : double, baseline : double);
+ParaLineMetrics(width : double, height : double, asc : double);
 OptimizedLineElement(f : GhostForm, metrics : Behaviour<FormMetrics>, interactivityIdM : Maybe<int>);
 
-getOptimizedLine(words : [ParaWord], alignment : ParaLineAlignment) -> [OptimizedLineElement]{
-	if (!isBiDiEnabled() && alignment != Justify()) {
+getOptimizedLine(words : [ParaWord], alignment : GeneralAlignment) -> [OptimizedLineElement]{
+	if (!isBiDiEnabled() && equalDoubles(alignment.justify, 0.0)) {
```
ykuzmenko73: really bad code here. why we have to compare double instead of just check if it justify?

lib/tropic/tropic2form.flow
```diff
@@ -28,6 +28,7 @@ makeContext(availableWH : Transform<WidthHeight>) -> TParentInfo {
 	TParentInfo(
 		fwidth(availableWH),
 		fheight(availableWH),
+		getDefaultRtl(),
```
ykuzmenko73: is it OK with new feature called TextDirection on the paragraph's level?

lib/form/paragraph3.flow
ykuzmenko73: Actually, I can't see any benefits from introducing GeneralAlignment. The code in general becomes more obfuscated from my point of view, more odd checks are required now.
Do we really need it? what are the benefits?

lib/form/paragraph3.flow
brawaga: My bad, I thought I pushed this.

lib/form/paragraph3.flow
brawaga: I have to agree.

lib/form/paragraph3.flow
brawaga: The odd check is about justify, I will correct that. Other is pure arithmetics, we have exactly the same for TAlign.

lib/form/paragraph3.flow
brawaga: I thought when the manager is defined, we need to use the behaviour from there. Do you suggest to connect it with a segregated behaviour I should create there anyway?

brawaga: I will correct.

lib/form/paragraph3.flow
brawaga: This is done once paragraph is initialized. I will look how to cut down the overhead calling the measuring only here.

lib/form/paragraph3.flow
brawaga: This was about cropping does not solve the issue, I wanted you to see this, and, maybe, commit until next PR ready which solves this issue.

lib/form/paragraph3.flow
brawaga: Now nowhere, so I ought to comment this as it is done for Available2/Size2.

lib/form/paragraph3.flow
brawaga: Bad habits from untyped languages. Sure.

lib/form/paragraph3.flow
brawaga: Sure, bool will bring more clearance.

lib/tropic/tropic2form.flow
brawaga: I will segregate this.

lib/form/paragraph3.flow
```diff
@@ -143,6 +168,9 @@ renderParagraph(
 	words : [ParaWord] = map(list2array(glued), makeParaWord);
 
 	aw = make(0.0);
+	width = if (widthManager == dummyWigiParagraphWidthManager) make(100000.0) else widthManager.width;
```
chekaninmaxim: looks like variables width, height and size do nothing. Why do the needed?

lib/form/paragraph3.flow
ykuzmenko73: aligning the code with tropic is good indeed

lib/form/paragraph3.flow
ykuzmenko73: sure we should use width from width manager if it is here. I'm talking about the constant 100000.0 you have places in many places in the code

lib/form/paragraph3.flow
ykuzmenko73: try to avoid double rendering. it is bad. we should not do it.

lib/form/paragraph3.flow
```diff
@@ -630,160 +674,155 @@ keepFormTogether(form : Form) -> bool {
...
-	interWord = if (alignment == Justify() && length(optimizedLine) > 1) {
-		nwords = length(optimizedLine);
-		remaining = (width - lineWidth);
-		min(10.0, remaining / i2d(nwords - 1));
-	} else 0.0;
-
+	nwords = length(optimizedLine);
+	interWord = if (nwords <= 1) 0.0 else
+		min(10.0, dWidth * alignment.justify / i2d(nwords - 1));
+	x = ref if (rtl) dWidth-startOffset-interWord*i2d(nwords-1) else startOffset;
```
chekaninmaxim: please, use spaces around arithmetic signs, it's hard to read such expressions

ykuzmenko73: and this expression is hard to understand in anyway. looks like a mess

lib/tropic/tropic.flow
```diff
@@ -512,14 +512,18 @@ TDynamicGroup(stackChanges : DynamicBehaviour<[TGroupChange]>) -> TDynamicGroup2
 		maxX : Transform<double>,
 		// How much is allocated to this element, including any y-fill expanders here?
 		maxY : Transform<double>,
-
+		// Text flow direction, it is convenient to use bool
+		// while two alternatives only, otherwise to replace
+		// with union everywhere, including ALL suffixed/prefixed
+		// with LTR/RTL.
+		isRTL : bool,
```
artxp: We don't use all capital letters, so it should be isRtl

lib/form/paragraph3.flow
brawaga: That is reserve for future PR. Ok, I'll comment it out for a while.

lib/form/paragraph3.flow
brawaga: I will.

lib/tropic/tropic.flow
brawaga: Renamed in segregated branch.

lib/form/paragraph3.flow
```diff
@@ -44,6 +44,7 @@ reconstructed on each reflow, because we combine them together to have as few sp
 possible, and because rendering words in the same sprite improves the interword spacing.
 */
 
+LARGE = 100000.0;
```
ykuzmenko73: bad naming. the name should describe the purposes of the variable/constant.
for example, paragraphWidthLarge looks better candidate

lib/form/paragraph3.flow
```diff
@@ -143,24 +153,36 @@ renderParagraph(
 	words : [ParaWord] = map(list2array(glued), makeParaWord);
 
 	aw = make(0.0);
+	width = if (widthManager == dummyWigiParagraphWidthManager) make(LARGE) else widthManager.width;
+	// height = make(0.0);
```
ykuzmenko73: remove commented code

lib/form/paragraph3.flow
```diff
@@ -112,7 +115,14 @@ renderParagraph(
 	widthManager : WigiParagraphWidthManager
 ) -> Form {
 
-	alignment : ParaLineAlignment = extractParaLineAlignment(s);
+	alignment : Maybe<double> = switch (extractParaLineAlignment(s)) {
```
ykuzmenko73: @alstrup , looks like we need your advice here
From my point of view, having union for alignment is better that compressing it into some tricky double value.
But @brawaga tells us that this is the common practice in the tropic world. 
If it is common and accepted then we agree to change paragraph to be like tropic. But maybe the change should be opposite - to make tropic be union-based?

lib/form/paragraph3.flow
```diff
@@ -143,24 +153,36 @@ renderParagraph(
...
+	width = if (widthManager == dummyWigiParagraphWidthManager) make(LARGE) else widthManager.width;
+	// height = make(0.0);
 	update = make(0);
 	linesB : DynamicBehaviour<[ParaLine]> = make([]);
 
-	renderParaLinesFn = \lines, storyWidth -> renderParaLines(
-		storyWidth, lines, tightWidth, alignment, interlineSpacing,
-		topLineBaseline, interactiveStyles, rtl, isParagraphMarked
-	);
+	optFn = \lines -> {
```
ykuzmenko73: optFn is not clear name. try to improve it

lib/form/paragraph3.flow
```diff
@@ -143,24 +153,36 @@ renderParagraph(
...
+		lasti = length(lines) - 1;
+		// None() for Justify, last line alignment overridden.
+		optzds = mapi(lines, \i, line -> getOptimizedLine(line, i != lasti && alignment==None()));
```
ykuzmenko73: optzds is used only once, why we need it at all?
anyway, it explain nothing because the name is not self-explaining

ykuzmenko73: I'm afraid it is copy-paste from renderParaLines, we should extract the common code into separate fn and use in both places

lib/form/paragraph3.flow
```diff
@@ -143,24 +153,36 @@ renderParagraph(
...
-	} else if (widthManager.infoOnly) {
-		// in this non-TightWidth case we just inform WidthManager about our width
-		// we will not take story width into account in this case, but use availableWidth as usual
+	renderedParagraph = if (widthManager == dummyWigiParagraphWidthManager || widthManager.infoOnly) {
 		Select(linesB, \lines -> {
 			rendered = renderParaLinesFn(lines, getValue(aw));
-			nextDistinct(widthManager.width, rendered.second);
-			rendered.first
+			nextDistinct(width, rendered.size.width);
+			// nextDistinct(height, rendered.size.height);
```
ykuzmenko73: remove commented code

lib/form/paragraph3.flow
```diff
@@ -170,8 +192,11 @@ renderParagraph(
 			);
 			// we have to notify story about changes in our width.
 			// it could alter widthManager.storyWidth if we are new winner so we have to defer
-			deferUntilNextFrameRendered(\->nextDistinct(widthManager.width, rendered.second));
-			rendered.first
+			deferUntilNextFrameRendered(\-> {
+				nextDistinct(width, rendered.size.width);
+				// nextDistinct(height, rendered.size.height);
```
ykuzmenko73: don't leave commented code in final PR please

lib/form/paragraph3.flow
```diff
@@ -630,160 +655,139 @@ keepFormTogether(form : Form) -> bool {
...
-	endOffset = if (rtl) 0. else max(0., width - lineWidth);
-	alignmentOffset = ref switch (alignment : ParaLineAlignment) {
-		LeftAlign(): if (rtl) endOffset else startOffset;
-		RightAlign(): if (rtl) startOffset else endOffset;
-		CenterAlign(): max(0.0, (width - lineWidth) / 2.0);
-		StartAlign(): startOffset;
-		EndAlign(): endOffset;
-		Justify(): 0.0;
-	} + (if (rtl) 0. else lineIndent);
+	dWidth = max(0.0, width - optimizedLine.metrics.width);
```
ykuzmenko73: please, use good naming convention.
dWidth and so on is not good naming

lib/form/paragraph3.flow
```diff
@@ -630,160 +655,139 @@ keepFormTogether(form : Form) -> bool {
...
+	startOffset = dWidth*either(alignment, 0.0) + optimizedLine.indent;
 	// Inter-word spacing for justification
-	interWord = if (alignment == Justify() && length(optimizedLine) > 1) {
-		nwords = length(optimizedLine);
-		remaining = (width - lineWidth);
-		min(10.0, remaining / i2d(nwords - 1));
-	} else 0.0;
-
+	nwords = length(optimizedLine.words);
```
ykuzmenko73: doesn't wordsCount look better?

lib/form/paragraph3.flow
brawaga: I did not tell it common in overall Ttropic, I told about TAlign only. Which works also in Material though. Also, this is internal only stuff, so I did not expect this to become a matter of conversation. When I did this, I tried to exclude multiple switches from code of repetitious execution up to paragraph initialization. Now I see maybe better to leave double instead Maybe<double> with exclusive (-1?) value for justify, but I am not sure about what can we get here: it does not seem to cut down if/switch quantity anymore, and I am unaware if we will get  some profit on parameter passing (union vs double).

lib/form/paragraph3.flow
brawaga: d is good for difference or delta. Why make the name twice longer while d represents the idea?

lib/form/paragraph3.flow
brawaga: That was here before and I never thought about, but I will rename if you like it like that.

lib/form/paragraph3.flow
brawaga: Ok.

lib/form/paragraph3.flow
ykuzmenko73: as for me, using distinct data members of union is a way better that tricky double values. it much easy to understand and produce less errors.
even if using double reduce couple of lines of code, it sacrifice understanding what is exactly is happens in the particular line of the code, require to lookup into the definition to understand the meaning of -1, 0, 0.5, 1
I'm not sure it worth it

ykuzmenko73: @brawaga ,
I can't see any improvements even on the code level from switching into double stuff. could you explain where it bring real benifits?

lib/form/paragraph3.flow
brawaga: Finally renamed as it was before.

lib/form/paragraph3.flow
```diff
@@ -44,6 +44,9 @@ reconstructed on each reflow, because we combine them together to have as few sp
 possible, and because rendering words in the same sprite improves the interword spacing.
 */
 
+INF_METRICS = 100000.0;
```
ykuzmenko73: We have several places where different "infinite" width are used (innovation repor).
They looks like 10000.0 or similar. Please, try to find those places and change to use this common constant instead

lib/form/paragraph3.flow
```diff
@@ -44,6 +44,9 @@ reconstructed on each reflow, because we combine them together to have as few sp
...
+
+NumAlignment ::= Some, Justify;
```
ykuzmenko73: Why Some is a part of NumAlignment? looks weird.

lib/form/paragraph3.flow
```diff
@@ -105,14 +110,29 @@ general2WrapElement(elem : GeneralWrapElement, style : [CharacterStyle]) -> Wrap
...
 renderParagraph(
 	p : [ParagraphElement],
 	s : [ParagraphStyle],
 	rtl : bool,
 	widthManager : WigiParagraphWidthManager
 ) -> Form {
 
-	alignment : ParaLineAlignment = extractParaLineAlignment(s);
+	alignment : NumAlignment = switch (extractParaLineAlignment(s)) {
+		StartAlign(): Some(0.0);
```
ykuzmenko73: I think it shows that the idea of double in align is not good.

lib/form/paragraph3.flow
brawaga: Only struct allowed as a part of union.

lib/form/paragraph3.flow
brawaga: I simply moved this switch from RenderLine to upper level. NumAlignment type just introduced internally to pass the value used underneath. It looks different because width remainder available in RenderLine only, so left on place.

lib/form/paragraph3.flow
brawaga: /lib/form/gui.flow line 403
InfiniteWidthAWH = const(WidthHeight(10000.0, 0.0))
?

PR #514, opened 2020-03-20
flowc: during helper server calls error messages are suppressed

dmitry-vlasov:


tools/flowc/flowc_lsp.flow
```diff
@@ -254,7 +254,8 @@ fcLspTextDocumentDocumentSymbol(call: FcLspMethodCall) -> void {
 					out = fcOutput(call.conf.threadId);
 					clearFcOutput(call.conf.threadId);
 					if (fcDetectLspError(out, call.conf)) {
-						call.out_cb(fcLspErrorResponse(call.id, 1, out));
+						//call.out_cb(fcLspErrorResponse(call.id, 1, out));
```
alstrup: Remove instead of uncommenting. Multiple times

dmitry-vlasov: done

PR #515, opened 2020-03-23
Flowc use decl if def is undef

dmitry-vlasov:


alstrup:
lgtm, although maybe a function could help?

PR #520, opened 2020-03-23
Moved all expression-related files into lib/wigi/expr

artxp:
There we have dependence of flow9 of innovation, what is wrong? These files should be probably moved somewhere in innovation repo.

### DIFF
--- a/lib/material/extra/calendar/calendar.flow
+++ b/lib/material/extra/calendar/calendar.flow
@@ -1,6 +1,6 @@
 import material/material_dialog;
 import material/material_ui;
-import wigi/excel_datetime;
+import wigi/expr/excel/excel_datetime;
 import text/translation;
 
 export {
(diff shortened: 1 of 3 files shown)


alstrup:
Surprising these rely on files from the innovation repo. That is against layering, since that means it is not self-contained in this repo. This PR is OK, but we should consider to move those files into innovation.

lib/material/extra/calendar/calendar.flow
```diff
@@ -1,6 +1,6 @@
 import material/material_dialog;
 import material/material_ui;
-import wigi/excel_datetime;
+import wigi/expr/excel/excel_datetime;
```
ykuzmenko73: Move lib/wigi/date_utils.flow from innovation into flow9 and use it in both places

alstrup: Agree, good solution.

PR #521, opened 2020-03-23
8739 picture custom zoom

kanakin-egor:
https://trello.com/c/CfFtJ9P9/8739-prime-graphics-used-in-probes-cannot-turn-off-zoom-feature

### DIFF
--- a/lib/ui/imagedecoratedstructs.flow
+++ b/lib/ui/imagedecoratedstructs.flow
@@ -1,6 +1,7 @@
 import stylestructs;
 import maybe;
 import form/form;
+import form/gui;
 import devmode;
 import form/paragraphtypes;
 import colorspace;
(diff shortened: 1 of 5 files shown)


lib/ui/imagedecoratedstructs.flow
```diff
@@ -54,8 +55,10 @@ export {
 		mouseZoom : bool,
 		extraZoom : bool,
 		closeButtonAlign : int,
-		closeButtonFn : (() -> void) -> Form
+		closeButtonFn : (() -> void) -> Form,
+		zoomButtonM : Maybe<ZoomButtonDescription>
```
ykuzmenko73: I would suggest to avoid Maybe here, create global constant defaultZoomButton and pass here if there is no custom

kanakin-egor: The problem is that standard zoom don't use button but cursor, so if we use default button stadard zoom behaviour will change.

PR #528, opened 2020-04-01
flowc: sync execution of a process native is added

dmitry-vlasov:


lib/sys/system.flow
```diff
@@ -11,6 +11,9 @@ export {
 	native setFileContentBinary : io (filename : string, content : string) -> bool  = Native.setFileContentBinary;
 	native setFileContentBytes : io (filename : string, content : string) -> bool  = Native.setFileContentBytes;
 
+	// Executes a process synchronously. Returns exit code.
```
alstrup, line 14: Add comment that it only works in Java.

dmitry-vlasov: done

PR #531, opened 2020-04-02
Flowc fix f12 for selected in vscode

dmitry-vlasov:


tools/flowc/flowc_helpers.flow
```diff
@@ -533,7 +533,10 @@ getNameByLineColFromFile(line : int, col : int, file : string) -> string {
 
 getNameByLineCol(line : int, col : int, src : string) -> string {
 	resolver = makeLineResolver(src);
-	getNameByPos(findLineIndex(resolver, line, col), src);
+	getter = \c -> getNameByPos(findLineIndex(resolver, line, c), src);
+	name = getter(col);
+	// Fix for vscode behavior: it may give a first column after the last symbol if a word was selected.
+	if (name != "") name else getter(col - 1);
```
alstrup, line 539: You should add a guard against col < 0 to avoid infinite loop, hu?

dmitry-vlasov: A guard is added

PR #535, opened 2020-04-06
C11 macro is enabled by default. Warnings are fixed

dmitry-vlasov:


platforms/common/cpp/core/STLHelpers.h
```diff
@@ -58,7 +58,7 @@ using std::flush;
 # define IOSCPP11 0
 #endif
 
-#if (defined(__cplusplus) && (__cplusplus != 1 && _cpluspus >= 201103L || IOSCPP11)) || defined(_MSC_VER)
+#if (defined(__cplusplus) && (__cplusplus != 1 && _cpluspus >= 201103L || IOSCPP11 || 1)) || defined(_MSC_VER)
```
alstrup, line 61: Why not simplify the condition accordingly instead?

dmitry-vlasov: ok, simplified

alstrup:
lgtm

PR #536, opened 2020-04-06
Add upload native file java native

dmitry-vlasov:


alstrup:
Why not just add a default flow implementation? Is this required?

dmitry-vlasov:
Serhii Mishchenko asked me to add a stub for this function. As I see, in order to implement it I need to implement a File native type and a bunch of corresponding functions from fs/filesystem.flow.

alstrup:
We doubt we need it implemented. My point is that you can add a default flow implementation, and then there is no need for the native. However, this is OK.

dmitry-vlasov:
It seems that this stub was put, but in the wrong native host: filesystem instead of http.

PR #541, opened 2020-04-07
Onboarding hint is shown above another dialog - need a callback when …

cs1924:
…dialog is closed or not shown


cs1924:
I need to know when the dialog is closed to show tutorials.
I can fix lyceum, is it safe for other repos?

NikitaFil:
LGTM. There is one more usage into Soccer. However, it has been already broken for a while.

PR #549, opened 2020-04-13
Dynamic size elements

cs1924:


cs1924:
Do you agree to add these shortcuts?

NikitaFil:
I`d suggest TSizedX, TSizedY for consistency with MFixed -> MFixedX/MFixedY

PR #554, opened 2020-04-17
Flowc specialization: fix bug with generic lambda

ykuzmenko73:
Current specialization code can't handle lambda with non-defined argument's type because it treats as generic but there is no code to handle generics in global lambdas

for example, simple test:
```
genericFn(arg : ?) -> string { "arg" }
lambda = \key -> { genericFn(key) }

main() {
    lambda("key");
    {}
}
```

compiles into

```
var A9__lambda=(function(key){
    return genericFn___c1(key);});
function main(){
  lambda___c1s("key");
    return null;
}
main();
```

in the branch it will compile into:
```
function A9__genericFn(arg){
  return "arg";
}
var A9__lambda___c1s=(function(key){
    return genericFn___c1s(key);});
function main(){
  A9__lambda___c1s("key");
    return null;
}
main();
```

it is not perfect fix because if some generic are used only inside lambda then the code will crash, but curator in the branch compiles and run seems correct because all generic functions mentioned inside lambdas are mentioned somewhere else in "normal" functions.

### DIFF
--- a/tools/flowc/flowc_version.flow
+++ b/tools/flowc/flowc_version.flow
@@ -2,5 +2,5 @@
 // Edit 'build-flowc' instead.
 export {
 	flowc_version = "3.0";
-	flowc_git_revision = "34fb520";
+	flowc_git_revision = "9370e14";
 }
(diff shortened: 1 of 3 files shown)


ykuzmenko73:
[flowc: implement in compiler specialization for cases like lookupTree](https://trello.com/c/UZiUhcyQ/728-flowc-implement-in-compiler-specialization-for-cases-like-lookuptree)

PR #556, opened 2020-04-20
Flowc completion fix

dmitry-vlasov:


tools/flowc/completion.flow
```diff
@@ -106,32 +106,32 @@ codeCompleteField(config : CompilerConfig, globEnv : FcTypeEnvGlobal, context :
 							mapping = fcLocalVars2types(toplevel, globEnv);
 							switch (lookupTree(mapping, context.fields[0].name)) {
 								Some(types): {
-									//fcDebugLog("codeCompleteField: X, types=" + i2s(sizeSet(types)), config);
+									fcDebugLog("codeCompleteField: X, types=" + i2s(sizeSet(types)), config);
 									foldSet(types, makeList(), \acc, type -> {
-										//fcDebugLog("type=" + prettyFiType(dummyPretty, type, makeSet()), config);
+										fcDebugLog("type=" + prettyFiType(dummyPretty, type, makeSet()), config);
```
alstrup, line 111: Should we enable all the debugging, or is that not required?

dmitry-vlasov: I already received some complains that struct field completion doesn't work and such debug info may reveal the reason why. In future, when this function will be stable debug info may be removed.

PR #563, opened 2020-04-21
fix 3d Java natives

Skifanya:


lib/fs/filesystem.flow
```diff
@@ -117,6 +117,10 @@ export {
 	isSamePath(p1 : string, p2 : string) -> bool;
 }
 
+createTempFile(name : string, content : string) -> native {
+	getFileByPath("Just some example");
+}
+
```
dmitry-vlasov, line 123: Not so good to return something clearly broken.
Maybe it's better to create a tmp file with a setFileContent function and some temporary name like "_tmp_1" ?
In case such file exists we can increment index until it is a new file.

Skifanya: Agree, fixed

PR #564, opened 2020-04-22
Flowc vscode windows issues

dmitry-vlasov:


RusArtm:
Now it doesn't remove symbols when ctrl+space

PR #567, opened 2020-04-24
MZoomToFit() fix for fillers

kanakin-egor:
https://trello.com/c/WswXBPhI/793-probe-stem-text-box-container-is-larger-than-the-screen-causes-horizontal-scrolling

### DIFF
--- a/lib/material/material_gui.flow
+++ b/lib/material/material_gui.flow
@@ -147,7 +147,6 @@ export {
 	// if the content fits in the box then it returns the content
 	// if the content does not fit then it returns a zoomed out content
 	// the final metrics is a metrics of the final content
-	// It may fail with filler as a box. Consider using MTweak([TFitZoom(true, true)], content, box) instead.
 	MZoomToFit(content : Material, box : Material) -> Material;
 	MScaleXY(factor : double) -> (Material) -> Material;
 	// Helpers for turning Material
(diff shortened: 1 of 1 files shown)


kanakin-egor:
@NikitaFil I tested it on real use cases in Rhapsode.
New-flag is for new features and not for fixes of existing problems AFAIK.
What am I supposed to write in the card? I fixed it but you should add new=1 to see it?

NikitaFil:
>  I tested it on real use cases in Rhapsode.

Then OK without a flag.

lib/material/material_gui.flow
```diff
@@ -487,21 +487,7 @@ MAvailableWidthHeight(m : Material, width : Transform<double>, height : Transfor
...
-			Factor(0.0, 0.0)
-		else
-			min(b.width / c.width, b.height / c.height)
-			|> (\fc -> Factor(fc, fc))
-	));
-
-	MGroup2(
-		MScale(scale, MInspect([ISize(cwh)], content)),
-		MFixSize(MInspect([ISize(bwh)], MVisible(make(false), box)), TEmpty())
-	)
```
kanakin-egor, line 504: @NikitaFil I'm a bit confused about this code. Is it ok to be deleted?
I don't see any difference in behaviour with and without it.

NikitaFil: I`ve expected from new implementation to fail in cases like `MLines2(MZoomToFit(...), MZoomToFit(...))`. 
But as it works fine both in separate testcases and lyceum usage, then ok to delete previous.

PR #569, opened 2020-04-27
vscode: servers may be started/stopped manually

dmitry-vlasov:


alstrup:
What is JS server?

dmitry-vlasov:
JS server is LSP protocol implementation written inside flow extension. There are two variants of LSP:   an old one, written by Michael Pliskin inside the flow extension, and the new one, native flow implementation. Now a user may switch to one of these implementations by commands.

alstrup:
Why have both?

dmitry-vlasov:
I left an old implementation for reliability. This redundancy may be removed in future, when the new LSP server will be tested enough. If you insist, I can remove the old implementation in this PR.

alstrup:
OK to keep it but call it legacy to help clarify. The new is default.

PR #581, opened 2020-05-06
Flowc repair wasm backend

dmitry-vlasov:


alstrup:
I believe you can just revert the jsmangling in the specialize code, since we did a copy of the specialization for JS instead. So just revert the mangling part, and let that be WASM only, and you do not need to thread the bool around.

dmitry-vlasov:
> I believe you can just revert the jsmangling in the specialize code, since we did a copy of the specialization for JS instead. So just revert the mangling part, and let that be WASM only, and you do not need to thread the bool around.

I looked at JS specialization - it uses functions like specializeFiFunctionDec from mangle.flow, so the mangling function mangleSuffix is uses in both versions: for JS and for WASM.

PR #583, opened 2020-05-07
Extended CharacterStyle to contain FSetDirection

artxp:
It was found out that the following code is incorrect 
https://github.com/area9innovation/innovation/blame/master/lib/wigi/wigify.flow#L1227, because CharacterStyle doesn't contain FSetDirection. This causes crashes in https://github.com/area9innovation/flow9/blob/master/lib/form/optimizeform.flow#L309

This change fixes this. Maybe it's worth to rename FSetDirection to SetDirection.


artxp:
paired with https://github.com/area9innovation/innovation/pull/1593

PR #584, opened 2020-05-08
Flowc move lsp types to formats

dmitry-vlasov:


lib/string_utils.flow
```diff
@@ -21,6 +21,12 @@ export {
 
 	// converts UTF32 into sequence of 16 bit chars according to https://www.unicode.org/faq/utf_bom.html
 	utf32toUtf16(c : int) -> string;
+
+	// Do a replacement "\n" to "\\n" and so on
+	escapeStr(s : string) -> string;
```
alstrup, line 26: Isn't this just "toString"?

dmitry-vlasov: It is very similar to 'toString', but it doesn't wrap the result with quotations

PR #587, opened 2020-05-15
Flowc java lib target

dmitry-vlasov:


alstrup:
Looks reasonable, but how do you use the result? Worth documenting in the readme.

dmitry-vlasov:
> Looks reasonable, but how do you use the result? Worth documenting in the readme.

Documentation on java backend, including usage of java-library option, is added to readme.

PR #589, opened 2020-05-20
Flowc repl

dmitry-vlasov:


dmitry-vlasov:
> Why flowc rather than flowc1?

flowc is just a string variable to hold the actual flowc program invocation. And it uses flowc1 compiler.

PR #591, opened 2020-05-22
Fixed Scheherazade standard font size is very small

artxp:
https://trello.com/c/Oaix9Pie/9259-scheherazade-standard-font-size-is-very-small-the-default-font-size-should-be-19-to-be-equivalent-to-tahoma-14

Before 
image

After 
image
image


artxp:
[Scheherazade standard font size is very small. the default font size should be 19 to be equivalent to Tahoma 14](https://trello.com/c/Oaix9Pie/9259-scheherazade-standard-font-size-is-very-small-the-default-font-size-should-be-19-to-be-equivalent-to-tahoma-14)

PR #592, opened 2020-05-26
Flowc java named main class

dmitry-vlasov:


alstrup:
As long as this does not break anything, LGTM.

dmitry-vlasov:
> As long as this does not break anything, LGTM.

Unfortunately it may break java building in some situations. In this PR I made a jar-building respect a full jar-option like 
jar=tools/fast/fastjar.  
In master the tools/fast/ prefix is swiped off, while in this PR it stays. So, this PR affects on where the produced jar file is placed.

alstrup:
Doubt that is a problem. OK

PR #597, opened 2020-05-28
added feature to support external sw modules (from components)

MalkoVl:
Fixes in the serviceWorker code.
Added feature to use custom code in the SW.


www/sw.min.js
```diff
@@ -1 +1 @@
...
...ta.urls.map(function(t){return function(t,r){var a=new Request(urlAddBaseLocation(t));return fetch(a).then(function(n){if(200!=n.status||"basic"!=n.type)return!1;var s=new Request(filterUrlParameters(a.url,r.map(function(e){return e.toLowerCase()})));caches.open(CACHE_NAME_DYNAMIC).then(function(r){return r.put(s,n.clone()),sendMessageToClient(e,{msg:"Cached resource:",url:t,urlCached:s.url}),!0}).catch(function(){return!1})}).catch(function(){return!1})}(t,e.data.data.ignoreParameterKeysOnCache)}))):"check_urls_in_cache"==e.data.action?(t=e.data.data.urls,Promise.all(t.map(function(e){return caches.match(urlAddBaseLocation(e),{ignoreSearch:!1}).then(function(t){return t?e:""}).catch(function(){return""})})).then(function(e){return e.filter(function(e){return""!=e})}).then(function(e){return{urls:e,status:"OK"}}).catch(function(){return{urls:[],status:"Failed"}})).then(r):"get_service_worker_version"==e.data.action?r({data:SERVICE_WORKER_VERSION}):"set_use_cache_only_in_offline"==e.data.action?(CacheMode.UseOnlyCacheInOffline=e.data.enabled,r({status:"OK"})):"get_requests_stats"==e.data.action?r({data:requestsCount}):r({status:"Failed",error:"Unknown operation: "+e.data.action})});
```
Eugeny48, line 1: Is this file still in use? Our applications use their own `sw.<app>.min.js`

MalkoVl: It is minified version of `base` sw file. Applications, which are generated by meta_app, it does not use it now. But may be it usefull for other applications.

PR #602, opened 2020-06-01
added parameter to hide spinner on image loading

MalkoVl:


MalkoVl:
@alstrup,  meda_block and wigi_decoration has sync code. Checking resources in the cache is async. So, we can get does we have image in the cache when the image will be procssed in the decoration code. Only behaviour can help us here...

alstrup:
OK

PR #610, opened 2020-06-12
MInputChips. Implemented.

NikitaFil:


artjom-vasin:
Screenshot 2020-06-14 at 13 04 55
I think it would be better to:
1. move the input field into the end, so chips will occupy all available width.
2. add an optional hint inside the input field such as "start typing to search"
3. add other styles of MAutocomplete() if it's possible.
And I don't understand why chips are invisible on MCard().

NikitaFil:
> 1. move the input field into the end, so chips will occupy all available width. 
Moved autocomplete into the grid.

NikitaFil:
> 2. add an optional hint inside the input field such as "start typing to search" 3. add other styles of MAutocomplete() if it's possible.

Added MCustomAutoCompleteStyle, so you can use MLabel now.

NikitaFil:
> And I don't understand why chips are invisible on MCard().

Fixed.

PR #613, opened 2020-06-17
json double with E parsing fixed

RusArtm:
There is a strict check for `e` but sometimes it is `E` and it works the same way.

Without this change 0E-8 will cause an error.


alstrup:
You might have to check whether s2d in all targets work with "e", but looks good.

RusArtm:
@alstrup the `e` character is already allowed. I just added `E`
Will check it for JS. Can't check other targets.

alstrup:
LGTM.

Coolrunner86:
It's fallback function, so JS target is not affected by that change. s2d recognizes both "e" and "E", checked it.

RusArtm:
Yes it does recognize `e` and `E` and seems to prodece correct result.
Tested with several cases. `e` and `E` produce the same result.
Merging...

PR #615, opened 2020-06-22
flowc: js: external natives are added

dmitry-vlasov:


alstrup:
LGTM. Probably worth adding a few lines to the readme explaining how to use this.

PR #619, opened 2020-06-26
Feature/753 fix html entities escaping issues

brawaga:
HTML overescaping/overunescaping fix.

Linked PR: https://git.area9lyceum.com/Lyceum/lyceum/pulls/1945
Card: https://trello.com/c/TGIsfVe7/753-fix-html-entities-escaping-issues


lib/form/paragraph3.flow
```diff
@@ -114,7 +114,11 @@ makeAlphabetAwareTextField(txt : string, style : [CharacterStyle]) -> native {
 	overridden = applyAlphabetStyles(detectAlphabet(txt), style);
 	textfield = makeTextfield(getMappedFontFamily(overridden));
 	dstyle = getDefinedTextStyle(overridden);
-	setTextAndStyle(textfield, txt, dstyle.face.family, dstyle.fontSize, dstyle.face.weight, dstyle.face.slope, dstyle.fillColor, dstyle.fillOpacity, dstyle.letterSpacing, dstyle.backgroundColor, dstyle.backgroundOpacity);
+	text = if (isUrlParameter753True()) {
```
ykuzmenko73, line 117: text is used once, inline it

lib/net/url_parameter.flow
```diff
@@ -45,6 +45,8 @@ export {
 
 	// Check if parameter given in URL
 	isUrlParameterSet(name : string) -> bool;
+
+	isUrlParameter753True() -> bool;
```
ykuzmenko73, line 49: This is bad. do not pollute library with strange and very local functions

lib/formats/xml.flow
```diff
@@ -188,7 +191,7 @@ makeTextParser(parameters : [XmlParsingOption]) -> (string, DynamicBehaviour<Xml
 		\s : string, events : DynamicBehaviour<XmlEvent> -> {
 			// Get rid of all CRs
 			s1 = strReplace(s, "\x0d", "");
-			next(events, XmlTextEvent(unescape(s1)));
+			next(events, XmlTextEvent(if (isUrlParameter753True()) s1 else unescape(s1)));
```
ykuzmenko73, line 194: The performance impact must be well measured because as I remember, the escaping using XML parsers was really slow - it tooks 15% of CPU in rendering wigi documents. measure it using large documents (10k size at least)

lib/form/paragraph3.flow
```diff
@@ -114,7 +114,11 @@ makeAlphabetAwareTextField(txt : string, style : [CharacterStyle]) -> native {
...
+		esc = extractStruct(style, EscapeHTML(true)).escape;
```
ykuzmenko73, line 118: if it is for CPP only then move extractStruct under cpp to avoid extra call in say JS

PR #625, opened 2020-06-30
MHintMarker. Implemented.

NikitaFil:


ykuzmenko73:
[Add new common property for wigi elements to turn on hint marker](https://trello.com/c/s6XKpY7b/10320-add-new-common-property-for-wigi-elements-to-turn-on-hint-marker)

PR #631, opened 2020-07-16
With feature fix

Coolrunner86:
Fixed multiple calls issue when source of "with" was a function call.


Coolrunner86:
> Does it only introduce a local if it is not already a variable? If so, OK

Is it important? I thought it's not and should not impact performance of so.

alstrup:
I think it matters but can come in a second improvement

Coolrunner86:
> I think it matters but can come in a second improvement

Done

alstrup:
ægtm

PR #633, opened 2020-07-21
Fs delete recursively is added

dmitry-vlasov:


alstrup:
No need for new flowc.jar

PR #638, opened 2020-08-03
padArray() fix

kanakin-egor:
padArray(), as stated in its description (https://github.com/area9innovation/flow9/blob/master/lib/ds/arrayutils.flow#L58), is supposed to fill the given array till the given length but instead it adds the given length to the initial length.

### DIFF
--- a/lib/ds/arrayutils.flow
+++ b/lib/ds/arrayutils.flow
@@ -312,22 +312,14 @@ foldi2(a : [?], init1 : ??, init2 : ???, fn : (int, ??, ???, ?) -> Pair<??, ???>
 }
 
 replaceAtAndPad(pad : ?, xs : [?], i : int, x : ?) -> [?] {
-	l = length(xs);
-
-	if (i < l) {
-		replace(xs, i, x)
-	} else if (i == l) {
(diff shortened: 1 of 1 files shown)


lib/ds/arrayutils.flow
```diff
@@ -312,22 +312,14 @@ foldi2(a : [?], init1 : ??, init2 : ???, fn : (int, ??, ???, ?) -> Pair<??, ???>
...
+	replace(padArray(xs, pad, i), i, x);
 }
 
 padArray(xs : [?], pad : ?, i : int) -> [?] {
-	if (i <= 0) {
+	if (i <= length(xs)) {
 		xs
 	} else {
-		padArray(arrayPush(xs, pad), pad, i - 1);
+		padArray(arrayPush(xs, pad), pad, i);
```
yegordovganich, line 322: logically looks correct, but I suggest you'd better use 
`concat(xs, arrayRepeat(pad, i - length(xs))`
not to call arrayPush in recursive

kanakin-egor: Thank you for the review!
Redid as requested.

PR #639, opened 2020-08-04
Feature/775 improve lib text compression lzwflow for memory and cpu consumption

ykuzmenko73:
JS:

```
 Org.size |Enc.t |Dec.t | Enc.size |%org|?
——————————+——————+——————+——————————+————+—
    239358|    41|    33|     45144| 18%|V
   2393580|   254|    44|    171680|  7%|V
   4787160|   400|    67|    258520|  5%|V
   7180740|   849|    76|    326628|  4%|V
   9574320|  1178|    97|    388588|  4%|V
  11967900|  1885|   121|    446260|  3%|V
  14361480|  2359|   167|    498548|  3%|V
  16755060|  3223|   160|    547056|  3%|V
  19148640|  3800|   198|    593404|  3%|V
  21542220|  4404|   180|    637988|  2%|V
  23935800|  4862|   189|    680788|  2%|V
```


lib/text/compression/lzw.flow
```diff
@@ -1,57 +1,75 @@
...
-			(if (strlen(^phrase) > 1) {
-				intAsString(lookupTreeDef(^dict, ^phrase, 0));
-			} else {
-				^phrase + zeroCode;
-			}),
-			^out
-		);
-	});
+	out = ref makeList();
+	phraseIdx = ref getCharCodeAt(s, 0);
```
brawaga, line 66: ref c

lib/text/compression/lzw.flow
```diff
@@ -1,57 +1,75 @@
...
+			};
+		};
+
+		c = getCharCodeAt(s, i);
+		switch (lookupIntTree(phrase, c)) {
+			Some(phi): {
+				phraseIdx := phi;
+			}
+			None(): {
+				code := ^code + 1;
```
brawaga, line 47: Any reason making code a ref? You are using it in a recursion.

brawaga:
I see you got rid of bitpacking? Maybe better to leave it via style or bool parameter, to extend range of future use. At least that did not influence compressing times in noticeable amount.
Also, hence memory usage is not the last criteria in the task you are solving at upper level, I'd better not deny opportunity to compress twice more than it is now.

lib/text/compression/lzw.flow
```diff
@@ -1,57 +1,75 @@
...
+				out := Cons(intAsString(^phraseIdx), ^out);
+				phraseIdx := c;
+			}
 		}
-		lzw_encode_rec(s, sLength, i + 1, code, phrase, dict, storeResult)
+
+		compressLzwRec(s, sLength, i + 1, dict, code, phraseIdx, out)
 	} else {
-		storeResult()
+		out := Cons(intAsString(^phraseIdx), ^out)
```
brawaga, line 56: Out looks like output value for the function. Maybe better to pass it also not as a ref?

lib/text/compression/lzw.flow
ykuzmenko73: indeed

lib/text/compression/lzw.flow
ykuzmenko73: moving to ref saves 20% in time

lib/text/compression/lzw.flow
ykuzmenko73: moving to ref saves 20% in time

ykuzmenko73:
> I see you got rid of bitpacking? Maybe better to leave it via style or bool parameter, to extend range of future use. At least that did not influence compressing times in noticeable amount.
> Also, hence memory usage is not the last criteria in the task you are solving at upper level, I'd better not deny opportunity to compress twice more than it is now.

bitpacking compression should be totally different function, not a style because the compressed data with bitpacking is not compatible with non-bitpacking algo.

lib/text/compression/lzw.flow
brawaga: Wow.

lib/text/compression/lzw.flow
brawaga: Wow.

lib/text/compression/lzw_bitpack.flow
```diff
@@ -0,0 +1,152 @@
...
+				phrase,
+				false
+			)
+		} else {
+			LZWDecodingAcc(acc with remainder=bitOr(acc.remainder, bitShl(c, acc.remainderLen)), remainderLen=16+acc.remainderLen)
+		}
+	});
+
+	list2string(Cons(out.phrase, out.list))
+}
```
brawaga, line 152: Didn't do any of same optimizations done for non-bitpacking ones?

brawaga:
> bitpacking compression should be totally different function, not a style because the compressed data with bitpacking is not compatible with non-bitpacking algo.

Even when same styles given, encoder and decoder were compatible though. But ok, this is Unix way, I like this too.

lib/text/compression/lzw.flow
ykuzmenko73: List mostly

PR #642, opened 2020-08-07
FontLoader: call StartFlowMain immediately after adding checked font into DOM

ykuzmenko73:
It reduces delay because of fonts from 3000-4000 ms down to 500-800

Paired: https://git.area9lyceum.com/Lyceum/lyceum/pulls/2117


alstrup:
Awesome

stanislavpodolia:
Noticed that on first load it can calculate wrong metrics for text (I guess it use Times New Roman as default). Plus in canvas it leaves default font if proper one wasn't ready before flow called. Added mtext_test.flow with example, clear cache to reproduce.

shkel:
image

it looks like the app is opening faster

ykuzmenko73:
[FontLoading: eliminate unwanted delay on application start due to fonts in some cases](https://trello.com/c/hcv6ttEg/10720-fontloading-eliminate-unwanted-delay-on-application-start-due-to-fonts-in-some-cases)

ykuzmenko73:
Instead of rewriting wbfont binding, I've solved issue with fonts.css / fontsconfig.json configuration.
now, it loads all fonts successfully without making them iactive

PR #645, opened 2020-08-17
Flowc fontconfig relative path

dmitry-vlasov:
Solution to: https://trello.com/c/CscV560c/735-flowc1-wrong-fontconfig-file-is-taken-in-in-server-mode


dmitry-vlasov:
> If it works, lgtm

I asked Yuriy Kuzmenko to test it - he made the trello card about the problem.

PR #646, opened 2020-08-17
Flowc js compare in brackets

dmitry-vlasov:
Fix to the bug: https://trello.com/c/cJwNMbwy/732-bug-in-js-interpreting-parenthesis


alstrup:
As is, this might have a real cost in terms of length of the generated JS code. Can we compare the size of curator.js with and without to assess this?

dmitry-vlasov:
> As is, this might have a real cost in terms of length of the generated JS code. Can we compare the size of curator.js with and without to assess this?

The difference of size for curator.js is 59137 bytes.

alstrup:
That is worth optimizing. Add a number for binding strength so we can add only those required. Can happen in a follow up. OK to merge this first.

PR #658, opened 2020-09-01
MScroll. MScrollSpinner. Added MCustomSpinner.

NikitaFil:


artxp:
image
looks similar to https://www.cssscript.com/ios-os-x-style-pure-css-loading-spinner/ to me

artxp:
@NikitaFil maybe we can add this kind of spinner as separate style?

PR #667, opened 2020-09-21
json: handle e in json parser

Skifanya:


lib/formats/json/json.lingo
```diff
@@ -19,10 +19,11 @@ jsonarray = "[" ws jsonvalue:v ("," ws ~jsonvalue:va {:va})*:vs "]" ws { consArr
 jsonbool = "true" ws { JsonBool(true()) }
 	| "false" ws { JsonBool(false()) };
 
-number = (int "." unsigned_int?)$s { s2d($s) } 
+number = (int "." unsigned_int? (exponent unsigned_int)?)$s { s2d($s) }
```
sfandersen, line 22: It should be: 
(int "." unsigned_int? ( ("+"|"-")? unsigned_int)?)$s { s2d($s) } 

I am not sure if s2d can handle the +, in that case just leave it out, but I think the "-" is important. 

See json syntax here https://json.org

PR #669, opened 2020-09-22
If java is not in the path, then give a warning

sfandersen:
This should make much easier to debug issues where Java is not in the path or if JAVA HOME is not defined. Before this the "flowcpp" and "update.bat" tools will give strange errors. Some thing like "(set not expected".


sfandersen:
This is committed directly in the master branch: https://github.com/area9innovation/flow9/commit/a80f86a44f43e4d5e50a41ea8d9c88794dc09f73

PR #672, opened 2020-09-28
Flowc use hashmap for speedup

dmitry-vlasov:


alstrup:
Can you check compile time against the old version?

PR #685, opened 2020-10-16
makeStyledTextfield: don't create garbage objects

ykuzmenko73:
Based on performance profiling of rendering complex wigi documents, the most expensive operation is garbade collection, it tooks 30%+ of the time so we should review all code and rewrite to not create garbage objects


ykuzmenko73:

NikitaFil (approved):
Looks good so far. And how much performance boost does it give?

ykuzmenko73:
it reduce load on GC, but still measuring the impact. GC is not stable beast :-(

PR #692, opened 2020-10-21
HaxeRuntime: tune compareByValue in JS

ykuzmenko73:
In JS there is better way to compare that haxe generates. it produces too many garbage.


ykuzmenko73:

alstrup:
In addition to objects (structs) and arrays, we also have refs and functions, which are also comparable. Are they still working as before?

ykuzmenko73:
@alstrup , curator starts fine and shows reader fine too so I hope it works as before....

ykuzmenko73:
The code was taken from the browser as generated code and put into untyped __js__ with small tweaks - garbage variables were removed, no other changes were maid

PR #694, opened 2020-10-21
HttpServerSupport.readInputStream tune performance and memory usage

ykuzmenko73:
Based on stackoverflow it is the best way
https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java

### DIFF
--- a/platforms/java/com/area9innovation/flow/HttpServerSupport.java
+++ b/platforms/java/com/area9innovation/flow/HttpServerSupport.java
@@ -184,8 +184,8 @@ public class HttpServerSupport extends NativeHost
 					result.write(buffer, 0, length);
 				}
 				return result.toString("UTF-8");
-			} catch (Exception e) {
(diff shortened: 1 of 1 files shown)


platforms/java/com/area9innovation/flow/HttpServerSupport.java
```diff
@@ -175,10 +175,18 @@ public void handle(HttpExchange exchange) throws IOException
 
 		private static String readInputStream(InputStream stream)
 		{
-			return
-				new BufferedReader(new InputStreamReader(stream))
```
sfandersen, line 179: Did this implementation also assume that the InputStream was UTF8?

ykuzmenko73: removed excplicit encoding to be the same as before - default charset will be used

PR #696, opened 2020-10-22
Flowc add memory cache info

dmitry-vlasov:


tools/common/config.flow
```diff
@@ -133,7 +110,9 @@ getFlowDir() -> string {
 			} else if (strContains(appPath, "bin/")) {
 				takeBefore(appPath, "bin/", appPath)
 			} else {
-				flowDirByFlowc();
+				println("WARNING: Could not determine the flow directory.");
```
ykuzmenko73, line 113: we have $FLOW environment variable set to the flow9 folder, why don't we use it here?

ykuzmenko73 (commented):
LGTM

PR #697, opened 2020-10-23
TextClip: tune setTextAndStyle in JS

ykuzmenko73:
Tuned:
image

Original:
image


ykuzmenko73:

PR #705, opened 2020-11-03
Implement grouping words which have same heading level under appropriate tags

chekaninmaxim:
…ate tags


chekaninmaxim:
[Accessibility. Make calculated cells to be under the same tag as heading text](https://trello.com/c/nrgzOmT9/814-accessibility-make-calculated-cells-to-be-under-the-same-tag-as-heading-text)

yegordovganich (approved):
it's too low-level, if it works then fine.

PR #710, opened 2020-11-09
MBorderT

PolazhinetsA:
BiDi-awareness is not the thing to apply when working with canvas


PolazhinetsA:
Found a better solution and don't need MBorderT anymore, but I'm leaving the PR open as it might be a welcome addition.

NikitaFil (approved):
Oh, it is a surprise that we still don't have it. Thanks for noticing.

PR #713, opened 2020-11-11
iOS: Implement metadata media stream scan for qr, etc. codes

vereschagai:


alstrup:
lgtm. Is there a JS api for this kind of thing as well for PWA? We would love to use QR codes in Rhapsode for login for very small kids.

vereschagai:
@alstrup That should be possible without big effort. Brief search and I found a library written in pure JS to determinate QR codes from ImageData, we grab from Camera: https://github.com/cozmo/jsQR.

alstrup:
Would be great with an example app that did it in JS, but this can be merged before then as far as I'm concerned. We need QR for login for kids shortly.

vereschagai:
> Would be great with an example app that did it in JS, but this can be merged before then as far as I'm concerned. We need QR for login for kids shortly.

The test js library app is here: https://cozmo.github.io/jsQR/.
OK, I am making JS implementation a priority.

PR #716, opened 2020-11-13
Fix return SQL Ints

sfandersen:
A session variable is a variable that begins with "@". For example:

CALL my_stored_procedure(1, @v);
SELECT @v;

If the stored procedure have an out variable that is of type int, then the type of @v will be bigint(20).

With out this fix the session variable is returned as SString. With this fix some of them are returned as SInt. 

I think there are a risk that this will break some code that depends on bigint being an SString.


Eugeny48:
The problem is that BIGINT (20 digits) may not fit in integer (10 digits). We should keep it as string.

sfandersen:
Ok. That make sense. I will look for another solution.

PR #717, opened 2020-11-20
MouseDownInfo: introduce insideY fn

ykuzmenko73:
https://trello.com/c/jT5a3uGa/11763-danfoss-cell-border-stroke-setting-in-the-table-properties-is-not-working-properly

In the WigiEditor we need an ability to check if the MouseDown happens inside Y coordinates but the default test for X is not suited because we need extended range on the left of the editor. so we need new insideY function to test

NOTE: it works in JS but crashes in CPP
image

crash in CPP:
image


ykuzmenko73:
[Danfoss: Cell border stroke setting in the table properties is not working properly](https://trello.com/c/jT5a3uGa/11763-danfoss-cell-border-stroke-setting-in-the-table-properties-is-not-working-properly)

PR #719, opened 2020-11-29
Synchronized funcs

dmitry-vlasov:


alstrup:
Add a comment about how to get a lock. It is not clear.

PR #722, opened 2020-12-04
Vscode webview

dmitry-vlasov:


tools/common/config.flow
```diff
@@ -388,14 +388,15 @@ jsCliParamsToTree(params : JSCliParams) -> Tree<string, string> {
 		Pair("isES6", b2s(params.isES6)),
 		Pair("isTypeScript", b2s(params.isTypeScript)),
 		Pair("readable", b2s(params.readable)),
-		Pair("beautify", b2s(params.beautify)),
```
alstrup, line 391: Are you changing the names of these parameters externally, or just internally in the compiler? I.e. will existing compile shell/batch script need updating?

dmitry-vlasov: Here I changed the names of properties to which come from the flowc_usage file. So I think the names which were used here were not correct.

PR #727, opened 2020-12-18
Fit long words feature 2.0

chekaninmaxim:
Apply scaling and reflowing of paragraph after too long word getting of too long word in paragraph.


chekaninmaxim:
[Fitting long words feature: Scale whole paragraph instead of only one long word](https://trello.com/c/UrXDxFNj/830-fitting-long-words-feature-scale-whole-paragraph-instead-of-only-one-long-word)

PR #731, opened 2020-12-23
Flowc catch fontconfig json errors

dmitry-vlasov:


dewshick (approved):

lib/formats/json/json_utils.flow
```diff
@@ -29,6 +32,9 @@ export {
 
 	// Prints a break down of the size of this piece of JSON
 	profileJsonSize(j : Json) -> void;
+
+	// Uses a JSON grammar and return None if input is not a valid JSON string.
```
alstrup, line 36: I would add this to a separate file, just to keep the dependencies small. json_parse.flow seems a good name.

dmitry-vlasov: Function parseJsonSafer is moved to json_parse.flow.

PR #732, opened 2020-12-23
Added an string to integer function that can fail

sfandersen:
Does this belong here?

Or should it be a local function.


alstrup:
I would say stringmath.flow instead. It is rarely used.

sfandersen:
I have move the function

PR #733, opened 2020-12-25
Flowc lsp fix stack overflow

dmitry-vlasov:


alstrup (approved):
If vsix is updated, remember version.

dmitry-vlasov:
> If vsix is updated, remember version.

flow.vsix is not changed.

PR #742, opened 2021-01-13
Consider some chinese chars as word dividers

chekaninmaxim:


chekaninmaxim:
[Use Chinese characters to split a string into words in wigi](https://trello.com/c/0raHcCMS/12379-use-chinese-characters-to-split-a-string-into-words-in-wigi)

lib/form/paragraph.flow
```diff
@@ -595,8 +595,8 @@ breakTextFragment(
 			if (isBiDiEnabled()) {
 				if (i >= ^chrCatPos) chrCatPos := remapMinusOne(getCharCatSplitPos(text, i+1, textLen)-1);
```
ykuzmenko73, line 596: @brawaga , take a look please. in the getCharCatSplitPos we are looking for some "breaks" so may be it should be generalized

chekaninmaxim:
@ykuzmenko73 Compared performance by 100 times measuring of  100 breakTextFragment calls with 6000 letters string.
avarage time is almost the same. Difference is in few miliseconds. So new behaviour does decrease performance

ykuzmenko73 (approved):

PR #746, opened 2021-01-27
Invalid fontconfig.json must fail build

brawaga:
JSON.parse fallback for C++ target made more strict and compatible, compiler now fails to build if fontconfig.json is invalid.


brawaga:
https://github.com/area9innovation/innovation/tree/feature/769-invalid-fontconfigjson-must-fail-build has linked changes.

lib/formats/json/json.flow
```diff
@@ -200,52 +199,61 @@ getJsonIntValueM(json : Json) -> Maybe<int> {
...
-			Pair(JsonArray(v.first), v.second);
+			Triple(JsonArray(v.first), v.second, v.third);
 		} else if (charCode == 123) {		// "{"
 			v = parseJsonObject2(json, jsonLength, start + 1, makeList());
-			Pair(JsonObject(v.first), v.second);
+			Triple(JsonObject(v.first), v.second, v.third);
 		} else if (charCode == 34) {		// "\""
 			v = deserializeRestOfString2(json, start + 1);
-			Pair(JsonString(v.first), v.second);
+			Triple(JsonString(v.first), start+1, v.second);
```
ykuzmenko73, line 216: Do we really need Triple?
If yes, Triple is bad choice here, the meaning of second and third is not clear. Seems special type is needed like ParseAcc

make clear comment in the code how and why it is used

lib/formats/json/json.flow
```diff
@@ -96,8 +96,7 @@ export {
 }
 
 parseJson(json : string) -> Json {
-	v = doParseJson(json, strlen(json), 0);
-	v.first;
+	parseJsonSafe(json)
```
alstrup, line 99: It seems we are just switching to the other one? Is that really the point?

lib/formats/json/json.flow
brawaga: I will rework this, thanks for your advice.

lib/formats/json/json.flow
brawaga: No, parseJsonSafe contains exactly those lines I removed, so nothing changes in means of functioning. In JS, this function is executed natively and the body defined is ignored anyway.

lib/formats/json/json.flow
brawaga: I will rework this, thanks for your advice.

lib/formats/json/json.flow
brawaga: No, parseJsonSafe contains exactly those lines I removed, so nothing changes in means of functioning. In JS, this function is executed natively and the body defined is ignored anyway.

PR #747, opened 2021-01-28
Vscode custom editor

dmitry-vlasov:


lib/sys/vscode.flow
```diff
@@ -0,0 +1,18 @@
+// A set of natives to communicate with vscode API. 
+// This set of natives makes sense only in JS backend.
+
+export {
+	// Send a message to a vscode. Usually it is a string which is a JSON object.
```
alstrup, line 5: Can you add a link to the VS code site where the messages are documented?

dmitry-vlasov: Links are added

PR #750, opened 2021-02-08
Wigi: make text metrics dynamic for Safari

ykuzmenko73:
In Safari, the text behaves like dynamic element regarding its metrics so we can't measure it in advance.
for example, the real metrics of the text is changes in case the scale is changed. but from pixi (canvas) point of view nothing changed.


ykuzmenko73:
[Praxis: Missing Spaces next to the bold words](https://trello.com/c/lyB3JQK5/12250-praxis-missing-spaces-next-to-the-bold-words)

alstrup:
If it works, OK with me. @stanislavpodolia Do you have any comments?

PR #756, opened 2021-02-16
JS: introduce FInputEventFilter

ykuzmenko73:
linked with https://github.com/area9innovation/innovation/pull/2112

in JS when pasting into FTextInput, we receive input both from paste and typing
but in WIGI we process pasting from global paste listener (special event Paste) so it leads to double processing of the pasting text.
this new filter allow to get type of event


ykuzmenko73:
[pasting on ipad in wigi is doubling things](https://trello.com/c/MzlOuHXe/843-pasting-on-ipad-in-wigi-is-doubling-things)

stanislavpodolia (approved):
Perfect!

PR #761, opened 2021-02-22
JS. FlowGraphics. Fixed positioning for strokeOpacity = 0.

NikitaFil:


mmissiato (approved):
Looks good in Prime but I can't tell much about changes.

stanislavpodolia (approved):
Tried to find a way to support invisible stroke without modifying graphicsBounds but it seems like the best solution the way you did it, good job!

PR #765, opened 2021-03-04
Support de-DU in getLang()

pzotov:
Task https://trello.com/c/ZiDXZh1I/12912-learner-landing-page-is-displayed-in-english-when-langde-du

### DIFF
--- a/lib/ui/fontmapping.flow
+++ b/lib/ui/fontmapping.flow
@@ -9,6 +9,7 @@ import fusion;
 import text/unicodecharacters;
 import net/urlhash;
 import fonts_repo;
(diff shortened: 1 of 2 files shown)


lib/ui/fontmapping.flow
```diff
@@ -565,16 +566,19 @@ getNativeLangFromBCP(bcpLang : string) -> string {
 	bcpLanguageList = [
 		Pair("es-mx", "la"),
```
pzotov, line 567: @kszh @dmiserg 
We change es-MX to la, but la was excluded from languages

PR #780, opened 2021-04-01
Fix surrogate pair utf82string

dmitry-vlasov:


alstrup:
Hard to review. Did you test?

dmitry-vlasov:
> Hard to review. Did you test?

Yes, I tested it with:
`import string;
main() {
println("💖 charcodes: " + strGlue(map(s2a("💖"), i2s), " "));
println("☢ charcodes: " + strGlue(map(s2a("☢"), i2s), " "));
println("€ charcodes: " + strGlue(map(s2a("€"), i2s), " "));
println("한 charcodes: " + strGlue(map(s2a("한"), i2s), " "));
println("𐍈 charcodes: " + strGlue(map(s2a("𐍈"), i2s), " "));
}`
The problem is that such a program (namely chars 💖  and 𐍈) causes a compiler crash because these chars are represented as a surrogate pairs and a compiler crashes at de-serializing the cache module for this program.

PR #788, opened 2021-04-15
Make foldiAsync2 tail recursive

yegordovganich:
It addresses "Maybe iterASync could be implemented simpler than using foldiAsync2" not to get stack overflow.


yegordovganich:
@MalkoVl can you please try to revert your https://github.com/area9innovation/flow9/commit/ee1174d02db4c2a15d687da77070e0f49d4daad3 and try your test case in this branch?

alstrup:
You are changing the semantics: Now it seems to visit in reverse order. That seems wrong.

yegordovganich:
> You are changing the semantics: Now it seems to visit in reverse order. That seems wrong.

Yes, it does in reverse but on purpose because it just call itself recursively and the first real fn call happens on the last array element. So it preserves natural order as it is now in master.

alstrup:
Then at minimum add a comment explaining how that happens.

PR #791, opened 2021-04-20
Flowc java gradle

dmitry-vlasov:


tools/common/config.flow
```diff
@@ -99,6 +99,15 @@ export {
 	// Gather includes from external flow.config's, listed in include_extern as CSV.
 	// When gathering external includes, relative paths are resolved to absolute.
 	fcExternIncludes(include_extern : string) -> [string];
+
+	// Tries to run a program with '-version' option, and output the result.
+	// In case a program doesn't exist on path or any other errors return an empty string.
+	fcProgramVersion(program : string) -> string;
+
+	runGradle(args : [string], currentWorkingDirectory : string, onStdOutLine : (out : string) -> void, onStdErr : (error : string) -> void, onExit : (errocode : int) -> void) -> native;
```
alstrup, line 107: This and fcGradleVersion  belongs somewhere in the java/ backend folder.

tools/common/pathutil.flow
```diff
@@ -130,3 +142,23 @@ fcCorrectUriPath(path : string) -> string {
...
+
+copyFileRelative(base : string, from : string, to : string) -> void {
+	if (!fileExists(from)) {
+		print("file: '" + from + "' doesn't exist - not possible to copy");
+	} else {
+		name = strRight(from, strlen(base));
+		path_to = pathCombine(to, name);
+		ensureDirectoryExists(dirName(path_to));
+		if (!copyFile(from, path_to)) {
+			println("error while copying file '" + path_to + "' from '" + from + "' to '" + to + "'");
```
alstrup, line 154: Why not change this to return a string with an error, so the caller can detect if it was done ok.

tools/flowc/flowc_usage.flow
```diff
@@ -54,6 +54,11 @@ printUsage(config) {
 	prih(help,"      java-main=<name>          Specify the java main class for the project. By default it is a name of a programm");
 	prih(help,"      java-incremental=1        Switch on incremental java target generation");
 	prih(help,"      java-library=id1,id2,...  Build a java library with id1, id2, ... interface");
+	prih(help,"      java-builder=<builder>    Use a specified build system. Now available: gradle or none");
```
alstrup, line 57: What is the default?

tools/flowc/flowc_usage.flow
```diff
@@ -54,6 +54,11 @@ printUsage(config) {
...
+	prih(help,"      java-incremental=1        Do not re-generate modules, which were not changed.");
+	prih(help,"      java-dependencies=<dep1,dep2,...>  External java libraries, used by java natives.");
```
alstrup, line 59: Is this for all build systems?

tools/common/config.flow
dmitry-vlasov: Moved to backends/java/fi2jar_utils.flow

tools/common/pathutil.flow
dmitry-vlasov: Copy functions have changed so that they return an error message / array of error messages

tools/flowc/flowc_usage.flow
dmitry-vlasov: Default is 'none' i.e. default built-in low-level jar builder. Comment is appended with this note.

tools/flowc/flowc_usage.flow
dmitry-vlasov: No, currently only for java. Potentially we could use npm for JS to fetch JS dependencies, but it seems to be quite different. Also there's a conan package manager for c++, but I'm not sure that it's needed now.

PR #812, opened 2021-06-07
Changed the positioning of the text-superscript

AlexandrRzz:
The superscript position is based on the height of the first character immediately before it without super/sub-script style.


ykuzmenko73 (commented):
LGTM

try to make ui test to control the metrics (it could be done using cursor's position)

PR #815, opened 2021-06-11
Flowc more verbose gradle

dmitry-vlasov:


alstrup (approved):
in general, looks fine. I noticed that when using jar=whatever.jar eitvout Gradle, it complains if Gradle is not found, but it works anyways. So that seems like a mistake. If Gradle is not used, it is not a problem that it is missing.

PR #816, opened 2021-06-11
tree: reverseTreeAnyway/reverseTreeExactly are added

dmitry-vlasov:


alstrup (commented):
I doubt these are common enough to have in the std. lib but maybe I'm mistaken? I would call the second one something with array in the same.

dmitry-vlasov:
> I doubt these are common enough to have in the std. lib but maybe I'm mistaken? I would call the second one so

I need such functions to reverse ids renaming (Tree<string, string>) - from long-shorter to shorter-long. 
Are reverseTree / reverseTreeArray names ok for these functions?

alstrup:
OK

fre. 11. jun. 2021 12.09 skrev Dmitry Vlasov ***@***.***>:

PR #820, opened 2021-06-17
Flowc java init modules

dmitry-vlasov:


alstrup:
What is the benefit of this?

dmitry-vlasov:
> What is the benefit of this?

It decreases to size of an implicit constructor of a class, which initializes all fields with default values. In rhapsode_server today there was a problem that it didn't compile by javac:
`/home/dmitry/area9/lyceum/components/rhapsode_server/javagen/rhapsode_server/rhapsode_server/rhapsode_server.java:2544: error: code too large
	public rhapsode_server(String[] args) {
	       ^
1 error`
This PR fixes the build

Skifanya:
LGTM

PR #836, opened 2021-07-16
Paragraph3: refactor reflowParaWords2 and fix wrong reflow

ykuzmenko73:
Currently, in reflowParaWords we measure every word and use it's width by summing it will availableWidth is overflow.
But resulting line is optimized so some texts are joined so the resulting width of the line might be different.
so if we set measured width as available width and call reflow again, it will change result.

in order to avoid such bad side effect, this PR uses the same measurement code as final width does.

this is very old bug, it is here from the very beginning. it leads to many bug and fixes, especially in wigiTable code.


lib/form/paragraph3.flow
```diff
@@ -255,6 +228,18 @@ renderParagraphEx(
 	renderedLinesInfoM : Maybe<DynamicBehaviour<[WigiRenderedLinesInfo]>>
 ) -> Form {
 
+
+	renderParagrapName = if (p == []) "" else {
```
brawaga, line 232: Typo.

brawaga (approved):
Looks nothing is lost, so fine.

ykuzmenko73:
[WigiTable - cell's content cuts if non-Start align is used](https://trello.com/c/uAmizDsh/870-wigitable-cells-content-cuts-if-non-start-align-is-used)

PR #838, opened 2021-07-26
Fixed `foldAsync` to avoid error "Maximum call stack size exceeded"

Eugeny48:


lib/ds/arrayutils.flow
```diff
@@ -240,7 +240,11 @@ foldiAsync2(xs : [?], i : int, acc : ??, fn : (int, ??, ?, (??) -> void, (???) -
 		onOK(acc)
 	} else {
 		fn(i, acc, xs[i], \acc2 -> {
-			foldiAsync2(xs, i + 1, acc2, fn, onOK, onError)
+			if (i % 64 == 63) {
+				deferred(\ -> foldiAsync2(xs, i + 1, acc2, fn, onOK, onError));
+			} else {
+				foldiAsync2(xs, i + 1, acc2, fn, onOK, onError);
```
yegordovganich, line 246: make recCall = \ -> foldiAsync2(xs, i + 1, acc2, fn, onOK, onError);

But what is the point? This "magic" definitely lacks some comment.

Eugeny48: timer breaks the chains of call, it starts a new chain

yegordovganich:
@alstrup what do you think about such a fix for foldAsync?
We've double checked it now as far as we got this trouble again, and it really fixes foldAsync issue.

alstrup:
In general, I'm OK with this, but can't we do more than 64 iterations before we have to defer?

Eugeny48:
Sure we can. How about 1000?

I took 64 from `allP`
https://github.com/area9innovation/flow9/blob/a3fa9913214e6848357958359d167e9210a074c2/lib/promise.flow#L180

PR #839, opened 2021-07-28
Wigi fix rendering

ykuzmenko73:


ykuzmenko73:
[WigiTable - in TightWidth, cell's content is not alway updated with actual availableWidth](https://trello.com/c/uAmizDsh/870-wigitable-in-tightwidth-cells-content-is-not-alway-updated-with-actual-availablewidth)

PR #861, opened 2021-09-09
Flowc java lambdas

dmitry-vlasov:


alstrup:
Looks good. Is there any performance difference measurable?

PR #867, opened 2021-09-24
JS: measureText fix bug with letterspacing

ykuzmenko73:
We have found that PIXI.measureText using canvas to measure text supports letterspacing only partly:
it counts it and adds to the measured width, but the canvas used for measuring does not aware about letterspacing at all so some fonts use ligatures so width measured by canvas is wrong in this case compared to the rendered in DOM real text.

in branch:
image

in master:
image

### DIFF
--- a/platforms/js/PixiWorkarounds.hx
+++ b/platforms/js/PixiWorkarounds.hx
@@ -613,8 +613,13 @@ class PixiWorkarounds {
 				nativeSetProperty.call(this, propertyName, value, priority);
 			}
 
-			PIXI.TextMetrics.measureText = function(text, style, wordWrap, canvas)
+			PIXI.TextMetrics.measureText = function(text0, style, wordWrap, canvas)
 			{
+				// We can't set letterspacing into Canvas so ligatures will affect measured width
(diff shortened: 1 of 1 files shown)


platforms/js/PixiWorkarounds.hx
```diff
@@ -672,7 +677,8 @@ class PixiWorkarounds {
 				{
 					let lineWidth;
 					lineWidth = widthContext.measureText(lines[i]).width / widthMulti;
-					lineWidth += (lines[i].length - 1) * style.letterSpacing + (style.wordSpacing ? style.wordSpacing * (lines[i].split(' ').length - 1) : 0.0);
+					lineWidth += (lines[i].length - 1) * style.letterSpacing / 2 + (style.wordSpacing ? style.wordSpacing * (lines[i].split(' ').length - 1) : 0.0);
```
NikitaFil, line 680: I'd say ((style.letterSpacing > 0) ? Math.floor(lines[i].length / 2) :  lines[i].length - 1) * style.letterSpacing is more accurate

PR #869, opened 2021-09-24
make correct error message and status

Skifanya:


platforms/java/com/area9innovation/flow/HttpSupport.java
```diff
@@ -213,12 +202,29 @@ public final Object httpCustomRequestWithTimeoutNative(String url, String method
...
 			}
 
-			onResponse.invoke(responseCode, response.toString(), responseHeaders.toArray());
-        } catch (MalformedURLException e) {
-        	onResponse.invoke(400, "Malformed url " + url + " " + e.getMessage(), new Object[0]);
-        } catch (IOException e) {
-        	onResponse.invoke(500, "IO exception " + url + " " + e.getMessage(), new Object[0]);
-        }
+			if (responseCode != 200) {
+				String errorMessage = con.getResponseMessage();
```
sfandersen, line 206: I think the getResponseMessage() will just return "Not Found" for all 404 responses.. 

See: https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#getResponseMessage()

platforms/java/com/area9innovation/flow/HttpSupport.java
```diff
@@ -213,12 +204,30 @@ public final Object httpCustomRequestWithTimeoutNative(String url, String method
...
+			} else {
+				inputStream = con.getErrorStream();
+			}
+
+			if (Objects.nonNull(inputStream)) {
+				String newLine = System.getProperty("line.separator");
+				BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
+				for (String line; (line = reader.readLine()) != null; ) {
+						response.append(line);
+						response.append(newLine);
```
sfandersen, line 221: For flow it should just always be "\n"... In case there is a configuration issue with java...

platforms/java/com/area9innovation/flow/HttpSupport.java
```diff
@@ -213,12 +204,30 @@ public final Object httpCustomRequestWithTimeoutNative(String url, String method
 				responseHeaders.add(kv);
 			}
 
+			InputStream inputStream = null;
+			StringBuilder response = new StringBuilder();
+
+			if (responseCode == 200) {
```
sfandersen, line 210: I would do: 
InputStream inputStream = con.getInputStream();
if (inputStream == null) {
   inputStream = con.getErrorStream();
}

sfandersen: 200 might not be the only none error response.

PR #872, opened 2021-10-01
Flowc java static runtime

dmitry-vlasov:


platforms/java/com/area9innovation/flow/Database.java
```diff
@@ -124,21 +124,17 @@ public RSObject(DBObject dbo, String query, Object[] queryParams) throws SQLExce
         }
     }
 
-	public void initializeDatabase() {
+	public void initialize() {
```
Eugeny48, line 127: We need `initializeDatabase` there https://github.com/area9innovation/innovation/blob/e7b73900bed09f1911f85031bf3c5dd49779cf62/platforms/java/com/area9innovation/flow/DatabaseSValue.java#L13

platforms/java/com/area9innovation/flow/Native.java
```diff
@@ -66,7 +66,12 @@ public Native() {
 		}
 	}
 
-	public final Object println(Object arg) {
+	public void initialize() {
+		Integer emptyList_id = FlowRuntime.struct_ids.get("EmptyList");
+		emptyList = FlowRuntime.struct_prototypes[emptyList_id];
```
Eugeny48, line 71: Why `EmptyList` is required? The flow program may not use Lists.

tools/flowc/backends/common.flow
```diff
@@ -179,6 +179,8 @@ export {
 		javaLibrary : Set<string>,
 		includes : [string],
 		debug : bool,
+		useLambdas : bool,        // Use lambdas instead of inner classes for closures
+		staticRuntime : bool,     // All methods in runtime are static
```
Eugeny48, line 183: Why do we need this? How can it do work?

Eugeny48: I mean native methods cannot be non-static now, right?

Eugeny48 (commented):
Has this already tested on MKS or RIS?

platforms/java/com/area9innovation/flow/Native.java
dmitry-vlasov: EmptyList is used in tailList native

tools/flowc/backends/common.flow
dmitry-vlasov: 1) No, I'll remove these config parameters.
2) yes, in static runtime all native methods must be static

dmitry-vlasov:
> Has this already tested on MKS or RIS?

No, it was tested only on compiler itself. To test on MKS or RIS we need to update native hosts in innovation and lyceum

platforms/java/com/area9innovation/flow/Native.java
Eugeny48: If  some my flow program do not use Lists, then `emptyList_id` will be -1, right?

dmitry-vlasov: I added a check on null value of emptyList_id.

PR #874, opened 2021-10-04
qt: main.cpp doesn't use QLatin1String.arg() since it is absent in Qt…

dmitry-vlasov:
… 5.12


RusArtm (commented):
Checked - works now!
Thanks!

RusArtm (approved):

PR #883, opened 2021-10-21
implement a base64 encryption/decryption for urls

vadlis:


sfandersen:
It looks good.

sfandersen:
It might be good with a commit message for why you changed the encryption code.

PR #898, opened 2021-11-08
flow9: supertype only for var=if

shkel:
example:
image

tests : 
image


tools/flow9/backends/cpp/cpp.flow
```diff
@@ -126,13 +126,13 @@ void flow_println(A d) {
 	}
 }
 
-bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
+bstatement2cpp(backend : Backend<CppInfo>, s : BStatement, parsedType : string) -> string {
```
alstrup, line 129: I don't understand what "parsedType" means. Is this the expected type, or something else?

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -142,15 +142,16 @@ bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
 						%else%
 					}",
 				[
-					"c", rece(cond),
-					"then", rec(then),
-					"else", rec(else_),
+					"c", rece(cond, getBType(cond)),
```
alstrup, line 145: The condition for an if is always Bool.

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -160,49 +161,84 @@ bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
 				""
 			} else {
 				"return ";
-			}) + bexp2cpp(backend, val) + ";"
+			}) + bexp2cpp(backend, val, type, "") + ";"
 		}
 		BTailCall(fn, args, values, type): {
-			fn  + "(" +superglue(values, rece, ", ") + ");"
+			fn  + "(" +superglue(values, \v -> rece(v, type), ", ") + ");"
```
alstrup, line 167: This is wrong, if it is expected type that is required. You have to get the type of each argument in the function type from the fn.

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -160,49 +161,84 @@ bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
...
+					"default", "default: " + rec(defaultCase.body, defaultCase.id) + ";",
+			]);
+		}
 	}
 }
 
 
-bexp2cpp(backend : Backend<CppInfo>, bexp : BExp) -> string {
-	rec = \ee -> bexp2cpp(backend, ee);
+bexp2cpp(backend : Backend<CppInfo>, bexp : BExp, resultType : BType, parsedType : string) -> string {
```
alstrup, line 209: Again, I'm not sure what parsedType is. Add a comment explaining this a bit better?

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +482,42 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
 		BTypeFunction(__, __) : makeTree(); // error
 	}
 }
+
+isBTypeUnionType(type : BType, backend : Backend<CppInfo>) -> bool {
+	switch (type : BType) {
+		BTypePar(id) : containsKeyTree(backend.info.unionTypes, id);
```
alstrup, line 488: This can never be a union type, so this is always false for BTypePar.

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +482,42 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
 		BTypeFunction(__, __) : makeTree(); // error
 	}
 }
+
+isBTypeUnionType(type : BType, backend : Backend<CppInfo>) -> bool {
```
alstrup, line 486: Why not just pass the CppInfo directly? Alternatively, move this to another file, in case it is more generally useful, and then let Backend<?> be polymorphic.

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +482,42 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
...
+	switch (type : BType) {
+		BTypePar(id) : containsKeyTree(backend.info.unionTypes, id);
+		BTypeName(id, __) : containsKeyTree(backend.info.unionTypes, id);
+		BTypeFunction(__, __) : false;
+	}
+}
+
+//up cast
+btype2superType(resultType : BType, type : BType, backend : Backend<CppInfo>) -> Maybe<string> {
```
alstrup, line 495: Explain what the args are. It is not clear to me which is the subtype (struct) and which is the supertype (union).

alstrup (commented):
Added some comments. I'm not sure it is exactly right as is. I do not understand why you need both a resultType and a parsedType, and I do not understand the meaning of these.

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +490,76 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
...
+	}
+
+	eitherMap(
+		superTypeId,
+		\id -> if (containsKeyTree(info.structIds, id)) upcast(id) else type,
+		type
+	);
+}
+
+btype2letValues(backend : Backend<CppInfo>, supertype : BType, type : BType) -> Pair<string, (string) -> string> {
```
alstrup, line 523: Add a comment explaining what this one does

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +490,76 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
...
+	varType = if (supertype == bvoidType || supertype == type) {
+		"auto"
+	} else {
+		btype2cpp(backend, supertype);
+	}
+	getStructTypeId = \structType : BType -> switch (structType : BType) {
```
alstrup, line 529: Why not just use a string, and let the empty string be missing a type id?

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +490,76 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
...
+		BTypePar(__) : None();
+		BTypeName(typeId, __) : if (containsKeyTree(backend.info.structIds, typeId)) Some(typeId) else None();
+		BTypeFunction(__, __) : None();
+	}
+
+	// TODO: flat
+	isSubStruct = \struct : BType, union : BType -> {
+		eitherFn(
```
alstrup, line 537: Use ?? instead of eitherFn, or rely on getStructTypeId giving a string

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +490,76 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
...
+			getStructTypeId(union),
+			\unionId -> eitherMap(
+				lookupTree(backend.info.unionTypes, unionId),
```
alstrup, line 540: lookupTreeDef?

tools/flow9/backends/bprogram.flow
```diff
@@ -29,7 +29,7 @@ export {
 	// Statement-based languages can get this format:
 	BStatement ::= BIfStatement, BLetStatement, BSetMutableStatement, BReturnVoid, BReturn, BTailCall, BSequenceStatement, BSwitchStatement;
 		BIfStatement(cond : BExp, then : BStatement, else_ : BStatement);
-		BLetStatement(id : string, value : BExp, body : BStatement);
+		BLetStatement(id : string, value : BExp, body : BStatement, type : BType);
```
alstrup, line 32: Add a comment explaining that this is the type of the id, which can be a supertype of the the value. Do we also need a type for the body, in case that needs a supertype cast?

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -445,3 +490,76 @@ switchType2caseIndexes(switchType : BType, backend : Backend<CppInfo>) -> Tree<s
 		BTypeFunction(__, __) : makeTree(); // error
 	}
 }
+
+liftStructureType(superType : BType, type : BType, info : CppInfo) -> BType {
```
alstrup, line 494: Add a comment explaining what this one does

tools/flow9/bexp/bexp.flow
```diff
@@ -12,7 +12,7 @@ export {
 		BVar, BLet, BIf, BCall, BLambda, BSequence, BConstruct, BField, BSetMutable, BSwitch, BArray;
 
 		BVar(id : string, type : BType);
-		BLet(id : string, value : BExp, body : BExp, type : BType);
+		BLet(id : string, value : BExp, body : BExp, type : BType, supertype : BType);
```
alstrup, line 15: Explain that this supertype is of the value, which type is for the body. Maybe do it like this?

   BLet(id : string, idType : BType, value : BExp, body : BExp, bodyType : BType);

alstrup (commented):
OK to merge as is, and we can refine from here.

PR #908, opened 2021-11-15
Flow9 : fold

shkel:
- native fold
- fix return in sequence
- fix capture

### DIFF
--- a/tools/flow9/backends/cpp/cpp.flow
+++ b/tools/flow9/backends/cpp/cpp.flow
@@ -106,7 +106,7 @@ bdeclaration2cpp(backend : Backend<CppInfo>, d : BDeclaration) -> string {
 						btype2cpp(backend, argtypes[i]) + " " + arg
 					}, ", "),
 					"rt", rt,
-					"body", bstatement2cpp(backend, body)
+					"body", bstatement2cpp(backend, body, true)
 				])
 			}
 		}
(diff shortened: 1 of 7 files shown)


tools/flow9/backends/cpp/cpp.flow
```diff
@@ -144,12 +152,12 @@ void flow_println(std::u16string d) {
...
 }
 
-bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
+bstatement2cpp(backend : Backend<CppInfo>, s : BStatement, withReturn : bool) -> string {
 	todo = \ -> {
 		println("TODO C++ statement: " + toString(s));
 		toString(s)
 	}
-	rec = \ss -> bstatement2cpp(backend, ss);
+	rec = \ss -> bstatement2cpp(backend, ss, true);
```
alstrup, line 160: withReturn

tools/flow9/backends/cpp/cpp.flow
```diff
@@ -168,13 +176,13 @@ bstatement2cpp(backend : Backend<CppInfo>, s : BStatement) -> string {
 		}
 		BLetStatement(id, value, body): {
 			"auto " + id + " = " + bexp2cpp(backend, value) + ";\n"
-			+ bstatement2cpp(backend, body);
+			+ bstatement2cpp(backend, body, true);
```
alstrup, line 179: withReturn instead of true

PR #914, opened 2021-11-18
Flow9 : fixes for c++

shkel:
- ':='
- 'return {};' and 'return (fn(); {});'
- error: template argument deduction/substitution failed (WIP)

image

### DIFF
--- a/tools/flow9/backends/cpp/cpp.flow
+++ b/tools/flow9/backends/cpp/cpp.flow
@@ -290,7 +290,7 @@ bstatement2cpp(backend : Backend<CppInfo>, s : BStatement, withReturn : bool) ->
 bexp2cpp(backend : Backend<CppInfo>, bexp : BExp, isGlobalExp : bool) -> string {
 	rec = \ee -> bexp2cpp(backend, ee, false);
 	switch (bexp) {
-		BVoid(): "{}";
+		BVoid(): "({})";
 		BBool(v): if (v) "true" else "false";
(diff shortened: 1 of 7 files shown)


tools/flow9/backends/cpp/cpp.flow
```diff
@@ -662,3 +666,48 @@ dType2id(type : DType) -> string {
...
+	fold(declarations, makeTree(), \acc, decl -> {
+		switch (decl) {
+			BGlobalVar(id, body, type): acc;
+			BStatFunction(id, args, body, tailCall, type): setTree(acc, id, type);
+			BExpNative(id, isio, type, nativeName): if (containsSet(nativesWhiteList, id)) setTree(acc, id, type) else acc;
+		}
+	});
+}
+
+bVar2realFuncTypes(var : BVar, backend : Backend<CppInfo>) -> string {
```
alstrup, line 688: Add a comment explaining what this does. It is not clear to me

tools/flow9/backends/type_hierarchy.flow
```diff
@@ -17,6 +17,9 @@ export {
 	);
 
 	extractPolymorphism(acc : Set<BTypePar>, b : BType) -> Set<BTypePar>;
+
+	// find a match between polyType (BTypePar) and real type (BTypeName)
```
alstrup, line 21: Add an example to clarify. I guess it is Pair<?, ??> against Pair<int, string> and you get ? -> int, ?? -> string?

PR #916, opened 2021-11-24
Refactoring color opacity

AlexandrRzz:
https://trello.com/c/7YH9PK1D/16264-colour-auto-adjusting-feature-for-table-borders
Related PR:
https://github.com/area9innovation/innovation/pull/2626
https://git.area9lyceum.com/Lyceum/lyceum/pulls/4093

### DIFF
--- a/lib/material/material_color.flow
+++ b/lib/material/material_color.flow
@@ -165,6 +165,18 @@ export {
 
 	findMColor(color : int) -> MColor;
 	findMColorWithBackground(color : int, background : int) -> MColor;
+
+	colorAddOpacity(color : int) -> int;
+
+	colorDeleteOpacity(color : int) -> int;
+}
+
+colorAddOpacity(color : int) -> int {
+	color + 0x7f000000;
+}
+
+colorDeleteOpacity(color : int) -> int {
+	color - 0x7f000000;
 }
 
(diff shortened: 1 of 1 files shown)


lib/material/material_color.flow
```diff
@@ -165,6 +165,18 @@ export {
...
+	colorAddOpacity(color : int) -> int;
+
+	colorDeleteOpacity(color : int) -> int;
+}
+
+colorAddOpacity(color : int) -> int {
+	color + 0x7f000000;
+}
+
+colorDeleteOpacity(color : int) -> int {
```
brawaga, line 178: To have this failproof, it is better to use bit operations, so bitAnd with 0xFFFFFF to remove and bitOr with 0x7F000000. Otherwise someone could add or remove twice leading to overflow/underflow and unexpected results.

brawaga (changes requested):
I suggested to make the function safer in code related comment.

brawaga (approved):

lib/material/material_color.flow
AlexandrRzz: https://github.com/area9innovation/flow9/pull/918

brawaga:
Agreed in skype to move bit masking to another PR or commit set.

PR #917, opened 2021-11-25
Flow9 : ref

shkel:
works correctly, but slowly (the reason was discussed).
image

@alstrup How about a solution to store a pointer to an object? I'm not good at c++. Maybe there is a nicer solution?

(I started an experiment with changing types to speed up here: https://github.com/area9innovation/flow9/tree/flow9_ref_v1)


alstrup (commented):
I guess we should use some ref. counted smart pointer instead.

shkel:
@alstrup  euler7 runs for 6 minutes with std::shared_ptr<> (maybe I used shared_ptr wrong)

PR #919, opened 2021-11-26
Flowc fix initorder

dmitry-vlasov:
Fix to the:
When a variable is mutually dependent on a function, i.e. there's no
strict dependence precedence between a variable and function,
if we put typechecking of a function before typechecking of a variable,
typecheker fails. So, in such situations we put all variables before
function in init order.
Appended:
When we resolve init order in a group of mutually dependent names:
  a) place all vars before functions
  b) order vars by dependency
  c) order funcs by dependency

https://trello.com/c/4GmNFWsZ/794-compiling-depends-on-naming


alstrup (approved):

Eugeny48 (approved):
It helped for my case

PR #926, opened 2021-12-07
qt: QtNatives are loaded by default

dmitry-vlasov:


alstrup (commented):
I can not assess this in detail, but it looks reasonable.

PR #927, opened 2021-12-08
MDynamicGroup. Added TCustomCombiner.

NikitaFil:


brawaga (approved):
Excellent. I haven't yet invented the API needed for TropicParagraph, but this way the code allows to amend the API easily. Probably, vertical and horizontal combiners are candidates to be a part of API, I am not sure.

PR #933, opened 2021-12-13
 small doc improvements for flow and vscode 

Qwsafex:
CharacterStyle is indeed a union, not a subtype. And I think full path to flow.vsix is clearer.


brawaga (approved):
LGTM.

PR #934, opened 2021-12-13
Flow9 : cast()

shkel:
image

image

@alstrup if an error occurs which option is better?
- exception
- random value (no errors in console)
- random value and an error


shkel:
image
image

tools/flow9/bexp/dexp2bexp.flow
```diff
@@ -129,7 +131,7 @@ dexp2bexp(env : BExpEnv, d : DExp) -> BExp {
 		}
 		DCast(value, target, pos): {
 			// We probably do not need this in the backends. It is purely a type checking thing
```
alstrup, line 133: Updste the comment above

alstrup (approved):

alstrup:
Exception is best if a flow cast is wrong.

PR #956, opened 2022-01-25
Fix bug when pebbles updated twice

kszh:
Task: https://trello.com/c/7nuhemmz/1405-discard-tablepage-parameter-during-the-search


bvaschenko (commented):
Could you please check MConfirmPebbleParameterSwitch and others before merge

PR #959, opened 2022-01-27
Fix bug with word spacing and annotations

AlexandrRzz:
https://trello.com/c/V5Spps48/16755-reader-works-slow-in-safari-if-text-is-justified?filter=label:Release%202.5
**Before:**
before
**After:**
after


ykuzmenko73:
image

lib/form/paragraph3.flow
```diff
@@ -1318,35 +1353,38 @@ optimizeLine(words : [ParaWord], alignment: ParaLineAlignment, availableWidth :
...
-				arr
-			}
 		} else {
 			arr
 		}
 	})
 }
 
+calcWordSpacing(words : [ParaWord], availableWidth : double, indent : double) -> double {
+			wordsCount = length(words) - 1;
```
ykuzmenko73, line 1375: extra tab?

ykuzmenko73: why do we need it as function? it is used only once, just inline the code

ykuzmenko73 (approved):

PR #972, opened 2022-02-11
lingo: prefix rules with a given prefix

dmitry-vlasov:
If two languages with the same rule names are used in one program, and
both have common rule names, like 'ws', 'digit', etc., and both
languages are compiled to flow, their AST mnemonics will contain same
globals like 'st_ws', 'st_digit' - which causes compilation errors.

Prefixing the rules with a unique prefix for each language solves this
issue.


lib/lingo/pegcode/pegcompiler.flow
```diff
@@ -467,3 +470,38 @@ removePegActions(pa : Parsing) -> Parsing {
 		LingoValue(value): pa;
 	}
 }
+
```
alstrup, line 473: Move these to a separate file. Maybe one of the existing helpers can be used to do this, but otherwise OK.

alstrup (approved):

PR #982, opened 2022-03-14
Lib wabt

dmitry-vlasov:


platforms/ts/wabt/WabtNativeHost.ts
```diff
@@ -0,0 +1,103 @@
...
+// Parses a WebAssembly text format source to a module.
+export function parseWat(filename: string, buffer: string, options: typenames.WasmFeatures): typenames.Promise<typenames.WasmModule, string> {
+	let fn = (fulfill: (x: any) => void, reject: (x: any) => void) =>
+		wabt().then(w => {
+			const m = w.parseWat(filename, buffer, options);
+			if (m) {
+				const module : typenames.WasmModule = {
+					name: "WasmModule",
+					_id : -1,
+					validate: () => m.validate(),
```
alstrup, line 13: The functions that can throw should be catched in this place, and we should return a bool or a string to map error handling back to flow.

lib/formats/wabt.flow
```diff
@@ -0,0 +1,90 @@
...
+    log: string
+);
+
+// Parses a WebAssembly text format source to a module.
+native parseWat : (filename: string, buffer: string, options: WasmFeatures) -> Promise<WasmModule, string> = WabtNativeHost.parseWat;
+
+// Reads a WebAssembly binary to a module.
+native readWasm : (buffer: [int], read_options: ReadWasmOptions,  options: WasmFeatures) -> Promise<WasmModule, string> = WabtNativeHost.readWasm;
+
+// Executes a WebAssembly function with given argiments.
```
alstrup, line 87: arguments with u

alstrup: Document what "flow" values really work here

tools/flowc/backends/common.flow
```diff
@@ -341,7 +341,7 @@ getFcBackendConfigs(config : CompilerConfig) {
 					"Wrong format of JS dependency: '" + dep + "'. Should look like one of the following:\n" +
 					"\tnpm[@braintree/sanitize-url;NativeHost]\n" + 
 					"\tyarn[@braintree/sanitize-url;NativeHost]\n" + 
-					"\tfile[/home/user/TS/lib.ts;NativeHost]", 
+					"\tfile[dir1/dir2/natives_sources.ts;NativeHost]",
```
alstrup, line 344: What is this path?

tools/flowc/tests/js/wabt1/readme.md
```diff
@@ -0,0 +1,24 @@
+Running the test
+----------------
+
+To run the test:
+
+    flowc1 nodejs=1 es6=test.js readable=1 wabt_test.flow
```
alstrup, line 6: Can we get it to work without "readable=1"?

lib/formats/wabt.flow
```diff
@@ -0,0 +1,90 @@
...
+	// Custom Annotation Syntax for the Wasm Text Format
+    annotations: bool,
+	// Garbage collection.
+    gc: bool
+);
+
+
+// A class representing a WebAssembly module.
+WasmModule(
+	// Validates the module. Throws if not valid.
```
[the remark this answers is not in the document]
dmitry-vlasov: Throws are caught and errors are exposed as non-empty strings with messages

platforms/ts/wabt/WabtNativeHost.ts
dmitry-vlasov: The 'validate' function is wrapped with try-catch and exceptions are caught and exposed as non-empty strings with messages.

lib/formats/wabt.flow
dmitry-vlasov: In theory following types: scalar, struct and array must work. I'm not sure about references. But it's not tested.

tools/flowc/backends/common.flow
dmitry-vlasov: It is an example path, maybe I should rather write 
`file[some_dir1/some_dir2/some_natives_sources.ts;NativeHost]`
to express that there's just a path to some .ts file

tools/flowc/tests/js/wabt1/readme.md
dmitry-vlasov: To make it working without 'readable=1' we need to access 'HaxeRuntime._structids_' in WabtNativeHost.ts - and yet I don't know how to do it.

PR #992, opened 2022-04-11
potential O(n^2) is removed from deserializeString

dmitry-vlasov:


alstrup (commented):
I don't get it. How could it be n^2?

dmitry-vlasov:
> I don't get it. How could it be n^2?

When " is met far away from the beginning of a large string, containing lots of \n and \t (i.e. a source file), the code
{code}
quote = strRangeIndexOf(s, "\"", start, strLength);
{code}
is invoked at each occurrence of \n or \t, and at each such occurrence the same position of far-away " is re-computed. 
I have met this with a large file (1.4 megabytes) being de-escaped in roughly 40 seconds, because there was very few of " symbols, but lots of new lines.

alstrup:
Add a comment explaining this, and then this is approved.

PR #997, opened 2022-04-14
WigiParagraph. Fixed dynamic words order in DOM.

NikitaFil:
https://trello.com/c/PeaYrVmP/17410-fitb-and-coach-have-an-illogical-reading-order-the-text-this-is-a-fill-in-the-activity-is-read-first-then-the-input-element-that


ykuzmenko73 (commented):
LGTM
@brawaga , please review too and check uitests

ykuzmenko73:
@NikitaFil , there is huge performance degradation 2-3 times
I've tested the branch with heavy reader - it's almost twice as slow (22 sec to SIMULATE instead of 13 seconds in master)
Closing SIMULATE (destroing the objects) a way slower too (45 sec instead of 13).
Entering into the editor much slower too (8.5 sec instead of 2.5)

brawaga (approved):
LGTM, except performance loss Yuriy discovered.

PR #1004, opened 2022-04-21
implement PBKDF2 hashing on java

vadlis:


bvaschenko (approved):
LGTM

platforms/java/com/area9innovation/flow/FlowPassword.java
```diff
@@ -0,0 +1,118 @@
...
+		byte[][] ret = {
+			correctpbkdf2,
+			requestedpbkdf2
+		};
+
+		return ret;
+	}
+
+	public static Boolean validateHash(String data, String hash) {
+		byte[][] comparables = getComparables(data, hash);
```
sfandersen, line 112: "password" would be a better name then "data". 

The getComparables is a very specific, it's code  can probably be included in the validateHash function, as it can be reused. Or just make the getComparables function private.

sfandersen:
I do not think you need the FlowPassword.hx file?

PR #1024, opened 2022-06-07
updated encode/decode for utf8 in cpp (behind compile flag)

MalkoVl:
UTF-8 is a base functionality for text in the application, but it works in different ways on different platform (checked only cpp and js). So, if we will encode specific text on the js target and decode it on cpp target - we will not get the original text. The same we got on cpp test for lyceum packages prepared on js.

More closely to original algorithm of encoding/decoding of UTF-8 is a cpp code. But it has two problems:
* there is no additional checks for special markers in the bytes on decoding (fixed);
* cpp used `unicode_char` type to store UTF-8 symbols (2 bytes type), but UTF-8 codes my be 3-bytes length. So, we can loos the text on decode/encode. It is not fixed and, seems, it is to much work to do that.

On js target historically becames that string symbols stored in 2 bytes type and all codes of 3 bytes length decoded (on the fly) in UTF-16 (2 bytes codes). So called, surrogate pairs. Native browsers knows that, detect that and knows how to process that. And it is in the browser/haxe native - we can fix it somehow (only to make special wrappers). So, on js we have:
* we use UTF-16 code string (not UTF-8);
* strlen, getCharCodeAt and other friends may works not correct because of surrogate pairs (1 original symbol coded with 2 surrogate symbols).

Good news:
* I added wrapper in cpp UTF-8 encode/decode to get the same string like it on the js target (encode/decode 3 bytes symbols into surrogate pairs). It is behind compiler flig `--use_utf8_js_style`;
* 3 bytes symbols we meet in the text not often and in most cases we do not feel this problem. An example of 3 bytes UTF-8 symbols is emoji.

With next commit will replace compile flag on app flag.

Also I working on js fixes - wrapper to use only UTF-8 coding (without surrogate pairs). Will push it with another PR.


platforms/common/cpp/core/STLHelpers.h
```diff
@@ -218,6 +218,11 @@ class IncrementalUtf8Parser
     bool is_complete() { return bytes == 0; }
 };
 
+// We store here compiler flag about which logic of utf8 decode we should use:
+//  false - usual way with loos real code for 3 bytes codes,
```
Eugeny48, line 222:
```suggestion
//  false - usual way with loss real code for 3 bytes codes,
```

platforms/common/cpp/core/Utf8.cpp
```diff
@@ -17,27 +17,89 @@
 #import <UIKit/UIKit.h>
 #endif
 
+// We store here compiler flag about which logic of utf8 decode we should use:
+//  false - usual way with loos real code for 3 bytes codes,
```
Eugeny48, line 21:
```suggestion
//  false - usual way with loss real code for 3 bytes codes,
```

NikitaFil (commented):
Have no remarks, LGTM.

platforms/common/cpp/core/Utf8.cpp
```diff
@@ -17,27 +17,89 @@
...
+//  true - with decoding 3 bytes codes into UTF-16 codes (pairs of 2 bytes codes).
+bool is_utf8_js_style = false;
+
 /* Character encoding */
 
-static inline int encode(uint8_t *out, uint16_t c) {
-    if (c <= 0x7F)
+static inline unsigned encode(uint16_t c, uint16_t next_c, bool is_last, uint8_t *out, int &cnt)
```
Eugeny48, line 27: It worth removing `inline`?

Eugeny48 (approved):
Where do these problems occur? On local PC when we run applications using byte runner?

platforms/common/cpp/core/Utf8.cpp
MalkoVl: Yes, `encode` function becomes a bit larger and `inline` looks not good now. Will remove it with next commit.

PR #1027, opened 2022-06-09
fixed file filtering by created user

emirbayramov:


lib/material/internal/filebrowser/material_filebrowser_internal_types.flow
```diff
@@ -22,6 +23,8 @@ export {
 			FbDropFileFn(fn : (files : [native]) -> void);
 			// allows to select several files at once
 			FbMultiSelect();
+			//filter file by created user
+			FbFileUserFilter();
```
MalkoVl, line 27: Recommended to add `userId` to support other cases of filter in future.

MalkoVl (commented):
Looks ok. Maybe we can extend this style a bit.

lib/material/internal/filebrowser/material_filebrowser_internal_types.flow
MalkoVl, line 27: Not sure about this one friend now. Is it used anywhere?

PR #1029, opened 2022-06-10
Allow to set url parameters from flow

pzotov:
Task https://trello.com/c/pMsCtuIp

Related changes in lyceum https://git.area9lyceum.com/Lyceum/lyceum/pulls/4931


yegordovganich:
I'm afraid a bit about side-effects but after discussing with Pavel, it seems we can add println in Rhapsode with warning that extra parameters set in URL thru the skin (and their values with devtrace). If @alstrup has no objections, let's merge.

PR #1036, opened 2022-06-29
Json any2json

dmitry-vlasov:


lib/formats/json/any2json.flow
```diff
@@ -0,0 +1,87 @@
...
+				[Pair("struct", JsonString(name))],
+				mapi(fields, \i, field ->
+					Pair(field, any2json(args[i]))
+				))
+			);
+		}
+	}
+}
+
+json2any(json : Json) -> flow {
```
Eugeny48, line 50: `json2any` does not produce refs, so the question is should `any2json` translate refs to values or to `JsonNull()`

Eugeny48 (approved):
LGTM

lib/formats/json/any2json.flow
dmitry-vlasov: thanks for noticing this moment, I think it may be configured with function params.

PR #1039, opened 2022-07-05
java: quit tries to wait until all threads are finished

dmitry-vlasov:


Eugeny48 (commented):
I think it will interfere with stopping the flowc server with "server-shutdown=1".

Do timers use the thread poll?

PR #1052, opened 2022-07-18
Fix "old" justify alignment for non breaking space and word joiner.

AlexandrRzz:


lib/form/paragraph3.flow
```diff
@@ -1141,6 +1152,18 @@ RenderLine(
 			} else fm
 		}
 
+		isSpaceForm = {
```
ykuzmenko73, line 1155: simplify and make function to test GhostForm, it looks like copy-paste now

ykuzmenko73 (commented):
small comments

ykuzmenko73 (approved):

PR #1054, opened 2022-07-25
Flowc add build time

dmitry-vlasov:


Eugeny48:
Flag `cpp2` does not work for me:
image

Will you add `flowc2` for Windows? How much better is `flowc2` than `flowc1`?

dmitry-vlasov:
> Flag `cpp2` does not work for me: image
> 
> Will you add `flowc2` for Windows? How much better is `flowc2` than `flowc1`?

cpp2 is a c++ target built on the base of QtByteRunner runtime. flowc2 is a flowc compiler, build with the cpp2 backend. Currently I use it as an alternative to flowc1, which doesn't depend on java backend. It is useful if I introduce significant changes of java backend, which affect runtime and compiler simultaneously so that I can't compile flowc with itself using java backend.

So, I'm not sure, that flowc2 is useful for any other purpose, except for mentioned above. Maybe it's better to exclude it from the PR.

PR #1057, opened 2022-08-02
JS: tune toString

ykuzmenko73:
It was found that toString implementation could be improved in JS, in some heavy usage scenario up to 3 times.

in master:
image

in branch:
image


ykuzmenko73:
[Educator: slow opening large modules for editing in assign action on IPad](https://trello.com/c/gTJ1pK9J/18290-educator-slow-opening-large-modules-for-editing-in-assign-action-on-ipad)

PR #1065, opened 2022-09-01
Do not start a new line if the current line is not filled yet

AlexandrRzz:
https://trello.com/c/qihacE94/2364-formatting-is-off


ykuzmenko73 (changes requested):
this is not stable solution because for example the width of recursion in the paragraph is not known in this time. so we can't rely on the width of element. we should introduce ZerroWidthElement and pass it like we do for Space

PR #1067, opened 2022-09-20
Flowc html no translate

dmitry-vlasov:


DudkevichD:
https://trello.com/c/4QaGoBQm/

DudkevichD:
Automatic translation is not working now, which is good!

tools/flowc/backends/common.flow
```diff
@@ -659,8 +660,9 @@ getFcBackendConfigs(config : CompilerConfig) {
 				htmlAdditionalScripts,
 				htmlPredefinedParams,
 				getConfigParameterDef(config.config, "working-dir", ""),
-				if (htmlManifest == "1") "manifest.json" else htmlManifest)
-			) else None();
+				if (htmlManifest == "1") "manifest.json" else htmlManifest,
+				getConfigParameterDef(config.config, "html-translate", "") == "1"
```
NikitaFil, line 664: Maybe it is better to name it "html-no-translate" to avoid confusion?

NikitaFil (approved):
Generally fine

tools/flowc/backends/common.flow
dmitry-vlasov: Option is renamed to 'html-no-translate', default value is 1.

PR #1068, opened 2022-09-22
sys: exception.flow is added

dmitry-vlasov:


Eugeny48:
I think it's unnecessary and even harmful in flow.

dmitry-vlasov:
> 

Can you please explain, why do you suppose exceptions unnecessary and even harmful?

>

Eugeny48:
Because flow is a functional programming language, functional languages avoid exceptions.

PR #1070, opened 2022-09-26
Replace ParaOneWord with ParaWordGlued

ykuzmenko73:
GlueTextFragments style was implemented using ParaOneWord  with defined Cols inside, it leaded to the broken interactive elements.
Replaced with ParaWordGlued - it instructs Word Grouping algorithm to keep words on the same line

### DIFF
--- a/lib/form/paragraph3.flow
+++ b/lib/form/paragraph3.flow
@@ -86,13 +86,13 @@ reconstructed on each reflow, because we combine them together to have as few sp
 possible, and because rendering words in the same sprite improves the interword spacing.
 */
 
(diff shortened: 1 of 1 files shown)


ykuzmenko73:
[MLP: tooltip is not working (randomly)](https://trello.com/c/uYfBoaI9/19200-mlp-tooltip-is-not-working-randomly)

brawaga:
The only things I suspect are switch removed from 622 line and the way new condition added to the expression at line 433.

AlexandrRzz (approved):
LGTM

PR #1076, opened 2022-10-05
Flowc js native hashmap haxe

dmitry-vlasov:


alstrup:
@theheadmen Hi, can you check whether this PR also gives the speedup? It is a more robust implementation of the hash map, but using a different approach. I suspect it might be slower.

PR #1079, opened 2022-10-07
hash: means to flexible use of FNV hashing are added

dmitry-vlasov:


lib/math/hash.flow
```diff
@@ -1,21 +1,30 @@
...
 	// Even faster version of hash for strings: uses only 'cutoff' symbols from the string
 	fastHashStringWithCutoff(key : string, cutoff : int) -> int;
+
+	// Constants for FNV hash: http://en.wikipedia.org/wiki/Fowler_Noll_Vo_hash
+	FNV_offset_basis = 0x811C9DC5;
+	FNV_prime = 16777619;
+
+	// Update hash with integer/double/string value
+	addHashInt(hash: int, v: int) -> int;
+	addHashDouble(hash: int, d: double) -> int;
```
alstrup, line 18: Add a comment that this allocates and is likely slow.

lib/math/hash.flow
```diff
@@ -48,9 +57,33 @@ fastHashStringTailRecursion(key : string, len : int, cutoff : int, acc : int, in
...
+	v3 = bitUshr(v2, 8);
+	bitXor(hash3, bitAnd(v3, 0xFF)) * FNV_prime;
+}
+
+addHashDouble(hash: int, d: double) -> int {
+	addHashBytes(hash, s2a(toBinary(flow(d))), 0);
+}
+
+addHashString(hash: int, s: string) -> int {
+	addHashBytes(hash, s2a(s), 0);
```
alstrup, line 82: Use getCharCodeAt and avoid s2a to do this hash without allocating in the heap.

dmitry-vlasov: Done.

lib/math/hash.flow
dmitry-vlasov: Done.

lib/math/hash.flow
```diff
@@ -1,21 +1,30 @@
...
 
 export {
 	hashString(key : string) -> int;
 	// Faster version of hash for strings, uses tail recursion
 	fastHashString(key : string) -> int;
 	// Even faster version of hash for strings: uses only 'cutoff' symbols from the string
 	fastHashStringWithCutoff(key : string, cutoff : int) -> int;
+
+	// Constants for FNV hash: http://en.wikipedia.org/wiki/Fowler_Noll_Vo_hash
+	FNV_offset_basis = 0x811C9DC5;
```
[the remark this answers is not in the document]
dmitry-vlasov: When a hash value is composed from components like int, double, string, etc. we need a value to start from. This value in FNV hash algorithm is FNV_offset_basis, thus we need to expose it in the module interface. Example of such hash:
`
LoGraphData : (
	nextLOId : int,
	previousLOid : int,
	depth : int,
	weight : double
);

loGraphData2hash(x: LoGraphData) -> int {
	hash1 = addHashInt(FNV_offset_basis, x.nextLOId);
	hash2 = addHashInt(hash1, x.previousLOid);
	hash3 = addHashInt(hash2, x.depth);
	addHashDouble(hash3, x.weight);
}
`

PR #1080, opened 2022-10-10
Ds intmap

dmitry-vlasov:


lib/ds/intmap.flow
```diff
@@ -0,0 +1,322 @@
...
+
+	// Make a copy of int map. Since IntMap is mutable, it may make sense.
+	copyIntMap(intMap : IntMap<?>) -> IntMap<?>;
+
+	equalsIntMap(h1 : IntMap<?>, h2 : IntMap<?>) -> bool;
+}
+
+// On java: wrapper around IntMap - has native implementation. NOT thread-safe. 
+//    About x2 of performance, compared with flow implementation (x10 compared with Tree).
+// On other platforms: flow implementation of int map.
```
alstrup, line 61: What about JS? We do have a fast one for JS, right?

PR #1082, opened 2022-10-11
Fill an icon in a tab the same color as text heading

MichelaSalvemini:
fill the icon by the same color as its heading when the "Color Selected Tab Text" is used, so the icon and text are in harmony

### DIFF
--- a/lib/material/internal/material_tabs.flow
+++ b/lib/material/internal/material_tabs.flow
@@ -141,6 +141,7 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
 						[onClick, AccessRole("tab"), FAccessAttribute("aria-controls", tabPanelId), MRippleType(const(MRippleFill()))]
 					);
 					tabIcon = tryExtractStruct(tb.style, MTabIcon(TEmpty()));
+					
 					rescaleEnabled = fand(rescaleEnabled0, extractStruct(tb.style, MTabTitleRescaleEnabled(const(true))).enabled);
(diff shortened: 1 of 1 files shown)


lib/material/internal/material_tabs.flow
```diff
@@ -145,27 +145,27 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
...
 					availableW = make(0.);
 					fullWidth = make(0.);
 					iconW = make(0.);
 					isIconVisible = make(true);
 					dynArrayPush(iconsVisible, isIconVisible);
 
 					TSelect2(tb.text, fand(const(!isKeepTabColor), feq(m.selected, i)), \txt, selected ->
-						if (txt != "") {
-							color = if (selected) Some(indicatorColor) else itemsColor;
+						TShow(const(txt != ""), {
```
NikitaFil, line 155: No need in using TShow here

lib/material/internal/material_tabs.flow
```diff
@@ -145,27 +145,27 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
...
+							if (isSome(iconAlign)) TFillX()
```
NikitaFil, line 156: Bug was introduced with the changed order of branches - text will disappear with MTabsIconAlign present.

NikitaFil (changes requested):
I'd add an additional MSelect for changed MIcon only. It may be helpful to extract 'color' from line #157 as a Transform and use it in this MSelect.

Notice also that this behavior should work only under MTabChangeIndicatorColor style to prevent already existing implementations.

PR #1083, opened 2022-10-12
ImageTextColumn decoration element [WIP]

Jamboba:


lib/ui/imagedecorated.flow
```diff
@@ -391,6 +393,51 @@ makeImageTextBox(ib : ImageTextBox) -> Form {
...
+	topLeft = Point(textCol.x, textCol.y);
+	fontSize = extractStruct(textCol.label.style, FontSize(36.)).size;
+	textColor = extractStruct(textCol.label.style, Fill(0xFFFFFF)).color;
+	isItalicFont = isItalic(textCol.label.style);
+	isBoldFont = isBold(textCol.label.style);
+	backColor = extractStruct(textCol.style, Fill(0));
+	backOpacity = extractStruct(textCol.style, FillOpacity(0.5));
+
+	material2nativeForm(
+		makeMaterialManager([]),
```
NikitaFil, line 408: Generally, using additional MaterialManager is a bad practice, but in this case it should be fine. In Material below manager is needed only for MParagraph transformation, but since custom color and font are used, it can be ignored. 

Trying to duplicate the same element in Form would be harder.

lib/material/internal/material_moving_point.flow
```diff
@@ -201,7 +203,7 @@ movingRectMaterial(
 			// Top Left corner
 			makeSubscribe(rectPointTLB, \rectPointTL -> {
 				if (curTLB |> getValue) {
-					limitX = getValue(rectPointBRB).x + iconSize;
+					limitX = getValue(rectPointBRB).x + iconSize;
```
brawaga, line 206: Extra space added.

lib/ui/imagedecorated_editor_types.flow
```diff
@@ -72,6 +72,20 @@ export {
...
 		);
+		ImageTextColumnEditable : (
+			x : DynamicBehaviour<double>,
+			y : DynamicBehaviour<double>,
+			width : DynamicBehaviour<double>,
+			fontsize : DynamicBehaviour<double>,
+			bold : DynamicBehaviour<bool>,
+			italic : DynamicBehaviour<bool>,
+			hAlign : DynamicBehaviour<CommonAlignment>,
+			vAlign : DynamicBehaviour<VerticalAlignment>,
```
brawaga, line 83: Maybe double 0…1 is better for VAlignment?

lib/ui/imagedecoratedstructs.flow
```diff
@@ -13,14 +13,15 @@ export {
...
-		PreOrPostZoomDecoration ::= ImageLabel, ImageCircle, ImageArrow, ImageRectangle, ImageTextBox;
+		PreOrPostZoomDecoration ::= ImageLabel, ImageCircle, ImageArrow, ImageRectangle, ImageTextBox, ImageTextColumn;
 
 		PreZoomDecoration ::= PreOrPostZoomDecoration, ImageDimmer; //Are grouped with full size image and then zoomed
 			ImageArrow   : (x : double, y : double, rotate : double, length : double, color : int, thickness : double, label : Form);
 			ImageCircle  : (x : double, y : double, radius : double, style : [GraphicsStyle], inverted : bool);
 			ImageLabel   : (x : double, y : double, label : Form); //Text is replaced with Form  as "Empty" can be found in serialized data
 			ImageRectangle : (x : double, y : double, width : double, height : double, style : [GraphicsStyle], inverted : bool);
 			ImageTextBox : (label : Text, hAlign : CommonAlignment, vAlign : VerticalAlignment, rectangle : ImageRectangle);
+			ImageTextColumn(label : Text, x : double, y : double, width : double, hAlign : CommonAlignment, vAlign : VerticalAlignment, style : [GraphicsStyle]);
```
brawaga, line 24: Also here, double is more flexible, and convenient in part where you adjust your text vertically. Single multiplication instead of cumbersome switch.

brawaga (changes requested):
A suggestion about vAlign. It is better and convenient to be type of double.

NikitaFil (approved):

brawaga (approved):
LGTM.

PR #1099, opened 2022-11-25
2432 introduce an option to animate panel and slide background images

MichelaSalvemini:
fixed TFillScale and added ease-in-out to MZoomAnimation


lib/material/measing.flow
```diff
@@ -698,7 +700,7 @@ MZoomAnimation(
 		],
 		MMEasingAnimation(
 			const(0.3),
-			const(easingFastOutSlowIn),
+			const(CubicBezierEasing(0.42, 0.0, 0.58, 1.0)),
```
mmissiato, line 703: Strange this constants in a component.
 @NikitaFil maybe you know better.

lib/tropic/tropic2form.flow
```diff
@@ -1281,7 +1281,9 @@ tropic2Acc(b : Tropic, parent : TParentInfo, sheet : Stylesheet, metricsOnly : b
 					xscale = fdivide(ba2Metrics.width, ba1.metrics.width);
 					yscale = fdivide(ba2Metrics.height, ba1.metrics.height);
 					customScale = extractStruct(style, TFillScale(const(0.0))).scale;
-					scale = fmax3(xscale, yscale, customScale);
+					scale =
```
mmissiato, line 1284: @NikitaFil please check this too.

NikitaFil: Don't change TFillScale behaviour, it will break existing content

lib/material/measing.flow
NikitaFil: It isn't a standard curve, but no objections about custom curve usage

NikitaFil (changes requested):
Are you sure you need MTweak here? Wouldn't it be easier to use just MScale?

NikitaFil:
@discane If you are fine with this new TFillScale behaviour, then I won't insist on changes. New naming / comment may be needed though.

PR #1101, opened 2022-12-01
Show tab header shadow on top of the tab background in preview

MichelaSalvemini:
Show tabs header shadow on top of the tab background in preview

### DIFF
--- a/lib/material/internal/material_tabs.flow
+++ b/lib/material/internal/material_tabs.flow
@@ -52,6 +52,17 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
 		default : acc;
 	});
 
+	headerOnTop = \tabHeader, tabsContent ->
+		TLet("tabHeader", TShadowShape(parent, elevation, tabHeader, BottomRectangleShape()), 
+			TGroup2(
+				TLines2(
+					TGhost("tabHeader"),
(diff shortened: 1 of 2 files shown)


lib/material/internal/material_tabs.flow
```diff
@@ -385,7 +396,6 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
 				)
 			)
 			|> (\t -> TFrame(0., 0., bgStyle, t))
-			|> (\t -> TShadowShape(p, elevation, t, BottomRectangleShape()))
```
NikitaFil, line 388: Why do we need to move the shadow from here?

lib/material/internal/material_tabs.flow
```diff
@@ -52,6 +52,17 @@ MDynamicTabs2T(manager : MaterialManager, parent : MFocusGroup, m : MDynamicTabs
 		default : acc;
 	});
 
+	headerOnTop = \tabHeader, tabsContent ->
```
NikitaFil, line 55: It looks like it is worth being a general-purpose element.
There is already MLines2InvertZorder. Let's make a copy of it, name TLines2InvertZorder, and locate in 'tropic-gui.flow'

NikitaFil (approved):

PR #1107, opened 2022-12-07
Flowc check unique import resolving

dmitry-vlasov:


Eugeny48 (commented):
It looks good, but sometimes applications have the same local file to import that has a higher priority  then the file from the innovation repo. So it is worth checking that the improvement will not harm our applications.

dmitry-vlasov:
> What are performance consequences for this? We do not want to scan all folders to check this, since that would too slow, especially on Windows.

The performance must be almost the same, unless a huge 'include' arrays of paths are in flow.config. The difference is: previous version scanned 'include' array until a first existent file is met, this PR scans all 'include' array to the end, finding all variants of existent combinations inclde + flowfile.

dmitry-vlasov:
> 



> It looks good, but sometimes applications have the same local file to import that has a higher priority then the file from the innovation repo. So it is worth checking that the improvement will not harm our applications.

For this case it's reasonable to make this check off by default, and turn it on explicitly with a compiler option

shkel:
I can't reproduce the bug :( Maybe it is related to the issue we have to clean up _objc_. But I don't have objections to PR

alstrup:
I am not comfortable with scanning all include folders for each file. That will probably be a slowdown, especially on Windows. We are compiling thousands and thousands of files, and checking file existance is super slow. Please check, or hide behind flag.

PR #1114, opened 2022-12-19
Introduced `setHttpRequestArgsUpdater` function

Eugeny48:
`setHttpRequestArgsUpdater` function can be used for:
1. Collecting requests (all, failed or successful);
2. Updating request parameters or headers;
3. Rejecting some requests;
4. Performing some requests on client side;

### DIFF
--- a/lib/net/http.flow
+++ b/lib/net/http.flow
@@ -1,17 +1,9 @@
 // © Copyright 2011 Area9 Technologies.
 
-import sys/target;
-import net/url;
-import ds/arrayutils;
 import fs/filesystem;
 import net/http_types;
 
 export {
-	/*currentRequest : DynamicBehaviour<Quadruple<string, bool, [KeyValue], [KeyValue]>> = make(Quadruple("", false, [], []));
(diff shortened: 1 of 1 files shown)


Eugeny48:
[Demo mode for Application](https://trello.com/c/NbKWgLj6/1509-demo-mode-for-application)

bvaschenko (commented):
Looks good to me. Just to be sure how it should be used: in case of demo mode we have to setHttpRequestArgsUpdater to
```
\args -> if (dbConnectorRequest(args) && readOnlyMode) {
    Some(args)
} else if (requestSupported(args)) {
    args.onData(preparedData);
    None();
} else {
    showRequestNotReadyWarning(args);
    None();
}
```
?
It looks like DW dbConnector requests are easy to distinguish by KeyValue("dbconnector", "true") parameter, what about MKS and interpreter dbConnector requests?

Eugeny48:
> in case of demo mode we have to setHttpRequestArgsUpdater to ...
Yes it's like that. MKS we can detect by URL and operation, and all RIS requests are read-only.

bvaschenko (approved):
LGTM

PR #1124, opened 2023-01-12
Fix svg scaling with Chrome

mmissiato:
- https://trello.com/c/Bp8dMc7w/2485-pngs-in-svg-containers-arent-scaled-properly-on-retina-displays

Before
Screenshot_20230112_113508
After
Screenshot_20230112_152644


NikitaFil (approved):
LGTM

PR #1127, opened 2023-01-16
20228 text problem in review mode

Seleng23:
https://trello.com/c/hjzjc6yg/20228-text-problem-in-review-mode


ykuzmenko73 (changes requested):
genIndent is dropped here, it will break a lot of content

PR #1130, opened 2023-01-17
Initial support for VS Code Dev containers

sfandersen:
Added support for dev containers in VS Code. See the devcontainer.markdown file for details.

The dev container still does not work with flowcpp. 

It should be possible to add support for flowcpp, but the asmjit git checkout inside the flow9 git checkout might be something people will need to do manually.


doc/devcontainer.markdown
```diff
@@ -0,0 +1,90 @@
...
+
+In VS Code with the DevContainer open, do the following in a Terminal: 
+
+	cd /workspaces/flow9
+	flowc1 jar=sandbox/hello.jar sandbox/hello.flow
+	java -jar sandbox/hello.jar
+
+If it is working, it should print: Hello console.
+
+Try the http target with:
```
alstrup, line 72: html target

alstrup (approved):
.md instead of .markdown is our convention.

PR #1135, opened 2023-01-23
unescape html

Skifanya:


lib/formats/xml.flow
```diff
@@ -176,6 +179,26 @@ xmlUnescape(s : string) -> string {
...
 }
 
+htmlUnescape(s : string) -> string {
+	xmlUnescape(s);
+	htmlUnescapeHelp(s);
+}
+
+htmlUnescapeHelp(s : string) -> string {
+	eitherMap(
+		findMatchingPairM(s, "&#", ";"),
```
sfandersen, line 189: You need to be able to handle: 
'&#x' [0-9a-fA-F]+ ';'   and 
'&#' [0-9]+ ';'

We should check that the text is a hex number or number, and only then try to convert the charater.



</body>
</html>

PR #1141, opened 2023-02-03
Improving VS Code extension

Eugeny48:
### Improved hover, now it works with variables in switches
image

with lambda's arguments
image

and with fields of structure
image

🐛 But it does not work with `with` statement. For example we have a variable `config : CompilerConfig` and use `with` statement to update field `config : Tree` of another structure.
image

### Improved F12
Declaration search works for local variables.


Eugeny48:
[Make F12 symbol look up find local functions](https://trello.com/c/vWP6OG1t/432-make-f12-symbol-look-up-find-local-functions)

alstrup (approved):

PR #1146, opened 2023-02-09
WIP Reworked mapList, mapiList and generateList to be tail-recursive

andrejefimov:


andrejefimov:
Pros:
+ stack space usage is O(1) instead of O(n); as a result it doesn't crash on Lists with 40k+ elements;
Cons:
- 25-50% slower than the old version; but this should not matter much on small samples.

andrejefimov:
filtermapList and filterList have less difference, new version can even be faster depending on the size and function

ykuzmenko73 (commented):
LGTM

PR #1151, opened 2023-02-22
fixed csv byte order mark

Qwsafex:


anastasiakucherova (changes requested):
I think we don't need to change the functions at all

PR #1165, opened 2023-03-13
for discussion : speed up getTreeValues

shkel:
curator_4

This PR speed up the fn, but there are 2 points to discuss here:
1) dlist2array is not good : dlist -> list -> array. (memory)
2) maybe a native Tree is better?

__
in PR:
dlist :
- make doDList2list tail recursive
- avoid garbage in doDList2list
- 
tree :
- attempt to improve getTreeValues

### DIFF
--- a/lib/ds/dlist.flow
+++ b/lib/ds/dlist.flow
@@ -166,19 +166,18 @@ popFirstDList(list : DList<?>) -> Maybe<?> {
 	switch (f : DNode) {
 		DEnd(): None();
 		DLink(v, __, __, __): {
-			removeDList(list, f);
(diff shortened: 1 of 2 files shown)


alstrup:
DList has bigger footprint than List? Why is this better? Worth understanding better

shkel:
> Why is this better? 

the reason is concat3
concat3(getTreeValues(left), [v], getTreeValues(right));

shkel:
i merged the part about dlist, because it is not relevant to the discussion.

alstrup:
Done in a simpler way using list2array and foldTree.

PR #1174, opened 2023-03-20
flowc: java: crashes are logged

dmitry-vlasov:


Eugeny48:
It would not work if the exception happens in the timer or any different thread.

Eugeny48:
The same can be done without fixing the compiler, just use a script with stderr redirected to a file.

PR #1175, opened 2023-03-20
flowc: JS: native implementation of a general hash map

dmitry-vlasov:


alstrup (approved):
Add comment explaining limits on what keys work in JS since structs for example do not, right?

PR #1179, opened 2023-03-30
Flowc nim comparison

dmitry-vlasov:
- general comparison operator is added. 
- pflow compiles, grammar is parsed successfully.
- some bugs in runtime/compiler are fixed


alstrup (approved):

alstrup (commented):
OK

PR #1185, opened 2023-04-13
Mfullscreen fix

mmissiato:
Fix for https://trello.com/c/4P1RD9mF/21026-ligth-editor-issue-click-on-help-with-make-all-other-buttons-unclickable
Related to https://trello.com/c/GiZvXaqj/20605-scorm-pacakge-in-preview-mode-if-you-zoom-in-scorm-package-will-dissappear
Can you please check if there is some regression in the related card?

### DIFF
--- a/lib/material/material_ui.flow
+++ b/lib/material/material_ui.flow
@@ -1813,7 +1813,7 @@ MFullScreenNoReattachments(fs : DynamicBehaviour<bool>, m : Material) -> Materia
 						false
 					)
 				),
-				false
+				true
 			)
 		);
 	});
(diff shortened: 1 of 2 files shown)


NikitaFil (changes requested):
appendToMainStage=false is the main characteristic of MFullScreenNoReattachments. With appendToMainStage=true it is just simple MFullScreen.

Buttons are still clickable, but because dialog is opened their onClick is blocked. The main issue is the fact that dialog is out of the requested fullscreen area, and it is hidden by browser.

We can't get rid of MFullScreenNoReattachments, it is required to prevent iframe from reloading. The only option I can see is to render help dialog, not as dialog but as a group inside fullscreen content.

mmissiato:
https://github.com/area9innovation/flow9/commit/6c96c573fa7c9dfd34f83b720c228d6e74429244

PR #1186, opened 2023-04-13
[Tooltip] New alignment style

MichelaSalvemini:
MBottomStart : align on botton right if rtl and bottom left otherwise

newTooltipStyle


NikitaFil (approved):
LGTM, but it would be better to add MTopStart, MTopEnd and MBottomEnd as well.

NikitaFil (approved):

PR #1192, opened 2023-04-25
iOS Capacitor: Implementation

vereschagai:


alstrup (approved):
Add docs to docs/runtime.md or similar about this.

PR #1203, opened 2023-05-10
Flowc java too long args

dmitry-vlasov:


tools/flowc/backends/java/fi2java_compile.flow
```diff
@@ -985,15 +995,13 @@ emitJavaExpression(ctx : JavaScopeContext, expcode : FiExp, rtype : string) -> s
...
-											type2javaObjType(ctx.func.gctx, sinfo.fields[j].type, true, false, false))),
-										", ") +
-									"))",
-									stype
+								sargs = strGlue(mapi(es, \j, e ->
+									emitJavaExpression(ctx, e,
+										type2javaObjType(ctx.func.gctx, sinfo.fields[j].type, true, false, false))),
+									", "
 								);
+								sargs1 = if (nfields < java_max_number_of_args) sargs else "new Object[]{" + sargs + "}";
```
shkel, line 1003: new Object[nfields]

shkel (commented):
ok, if it works

tools/flowc/backends/java/fi2java_compile.flow
dmitry-vlasov: I tried `new Object[nfields]` but this way of initialization fails:
`error: array creation with both dimension expression and initialization is illegal`

PR #1207, opened 2023-05-24
add style for changing content name

SLemishko:
https://trello.com/c/wybrNHsO/21539-answer-in-cat-probe-as-part-of-sequence-vanished-after-save-deploy


bvaschenko (changes requested):
this style is only for Flowschema, so shouldn't be there

PR #1210, opened 2023-05-31
Flowc nim fixes 3

dmitry-vlasov:
Utf16-based string is added, so the kind of string may be chosen by options. Utf16 is default, Utf8 is switched on by nim-utf8=1 option.


tools/flowc/backends/nim/runtime.nim
```diff
@@ -8,68 +8,341 @@ import hashes
...
-    else: r.add(ch)
-  return r
+#[ Flow string: either a UTF-16 string of UTF-8 ]#
+
+const use16BitString = when defined UTF8: false else: true
+
+when use16BitString:
+  type String = seq[Utf16Char]
+else:
+  type String = string
```
sfandersen, line 25: Just uppercase String is a quite generic name. Maybe "RtString" would be a better name.

tools/flowc/backends/nim/natives/string/toUpperCase.nim
```diff
@@ -1,2 +1,22 @@
-proc $F_0(toUpperCase)*(s: string): string =
-  return unicode.toUpper(s)
+proc $F_0(toUpperCase)*(s: String): String =
```
sfandersen, line 1: This will only convert English letters to Upper case. Maybe just convert to native strings and then use the native toUpper() function, and convert that back to utf16

tools/flowc/backends/nim/runtime.nim
```diff
@@ -8,68 +8,341 @@ import hashes
...
+  when use16BitString:
+    sequtils.concat(s1, s2)
+  else:
+    s1 & s2
+
+template rt_string_len*(s: String): int32 =
+  when use16BitString: int32(s.len) else: int32(s.runeLen)
+
+template rt_string_char_code*(s: String, i: int32): int32 =
+  when use16BitString: int32(cast[uint16](s[i])) else: int32(s.runeAtPos(i))
```
sfandersen, line 55: I think for utf8 strings the flow getCharCodeAt function should just return the 8bit value at the index (including surrogates). 

This way it is more like utf16 where it also return surrogates. I suspect this will make more flow code work.

tools/flowc/backends/nim/natives/string/fromCharCode.nim
```diff
@@ -1,4 +1,7 @@
 import unicode
 
-func $F_0(fromCharCode)*(code: int32): string =
-  return unicode.toUTF8(cast[Rune](code))
\ No newline at end of file
+func $F_0(fromCharCode)*(code: int32): String =
+  when use16BitString:
+    return @[Utf16Char(cast[int16](code))]
+  else:
+    return unicode.toUTF8(cast[Rune](code))
```
sfandersen, line 7: In Java the implementation uses the 32 bit code to make a uft16 string (so it can end up with a string of length 2)
In Cpp the implementation uses just cast to 16 bit, and use that as an single string of lenght 1.

This implementation does a mix of this. When use16BitString you only use the 16 bit, when not use16BitString you uses all the 32 bits. 

I suggest: 

when use16BitString:
    return rt_utf8_to_string(unicode.toUTF8(cast[Rune](code)));
  else:
    return unicode.toUTF8(cast[Rune](code))

sfandersen: An relevant optimization would be when use16BitString and "code & 0xffff0000 = 0" then we can use the fast way: 
 @[Utf16Char(cast[int16](code))]

tools/flowc/backends/nim/runtime.nim
dmitry-vlasov: Ok, agree, String is renamed to RtString.

tools/flowc/backends/nim/natives/string/toUpperCase.nim
dmitry-vlasov: Ok, switched to unicode.toUpper/toLower wrappers

tools/flowc/backends/nim/natives/string/fromCharCode.nim
dmitry-vlasov: Yes, on surrogate pairs this code will fail. I replaced with:
`    if (code < 0x0D800 or (0x0DFFF < code and code < 0xFFFF)):
      return @[Utf16Char(cast[int16](code))]
    else:
      # must be represented as a surrogate pair
      return rt_utf8_to_string(unicode.toUTF8(cast[Rune](code)))`

PR #1211, opened 2023-06-01
flowc: nim: utf-8 string indexes are byte-wise

dmitry-vlasov:


tools/flowc/backends/nim/runtime.nim
```diff
@@ -49,13 +49,10 @@ template rt_concat_strings*(s1: RtString, s2: RtString): RtString =
...
 template rt_string_len*(s: RtString): int32 =
-  when use16BitString: int32(s.len) else: int32(s.runeLen)
+  when use16BitString: int32(s.len) else: int32(s.len)
 
 template rt_string_char_code*(s: RtString, i: int32): int32 =
-  when use16BitString: int32(cast[uint16](s[i])) else: int32(s.runeAtPos(i))
-
-template rt_char_code_to_string*(code: int32): RtString =
-  when use16BitString: @[Utf16Char(cast[int16](code))] else: unicode.toUTF8(code)
+  when use16BitString: int32(cast[uint16](s[i])) else: int32(s.runeAt(i))
```
sfandersen, line 55: When the rt_char_code_to_string is used to implement the flow getCharCodeAt then for "use16BitString" is it does look like it always return a string with one uint16. It it should sometime be two words (like in the Java native)

tools/flowc/backends/nim/runtime.nim
```diff
@@ -479,7 +475,12 @@ proc rt_to_string_quot*(f: Flow): RtString =
   of rtVoid:   return rt_to_string()
   of rtBool:   return rt_to_string(f.bool_v)
   of rtInt:    return rt_to_string(f.int_v)
-  of rtDouble: return rt_to_string(f.double_v)
+  of rtDouble:
+    # NOTE: toString(42.0) == "42.0" BUT cast(42.0: double -> string) == "42"
+    var x = formatFloat(f.double_v)
+    x.trimZeros()
+    if not x.contains('.'): x.add(".0")
```
sfandersen, line 482: I think it should be if it does not contain "." AND not contain "e".

tools/flowc/backends/nim/natives/string/getCharAt.nim
```diff
@@ -1,13 +1,7 @@
 import unicode
 
 func $F_0(getCharAt)*(s: RtString, i: int32): RtString =
-  when use16BitString:
-    if i >= 0 and i < rt_string_len(s):
-      return @[s[i]]
-    else:
-      return rt_empty_string()
+  if i >= 0 and i < rt_string_len(s):
+    return when use16BitString: @[s[i]] else: unicode.toUTF8(unicode.runeAt(s, i))
```
sfandersen, line 5: It still seems wrong to use "rune" for the uft8 case. But Maybe we should just focus on utf16 first and get that working.

tools/flowc/backends/nim/runtime.nim
dmitry-vlasov: It happes that to check just for '.' is sufficient. Citation: "The exponential ("E") format specifier converts a number to a string of the form "-d.ddd…E+ddd" or "-d.ddd…e+ddd", where each "d" indicates a digit (0-9). The string starts with a minus sign if the number is negative. Exactly one digit always precedes the decimal point." So, in case of 'e' sign the dot must be also in the string.

tools/flowc/backends/nim/runtime.nim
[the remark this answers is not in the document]
dmitry-vlasov: Ok, fixed.

tools/flowc/backends/nim/natives/string/getCharAt.nim
dmitry-vlasov: ok, I switched to `return when use16BitString: @[s[i]] else: $s[i]`

PR #1214, opened 2023-06-08
Wigi: introduce selection support with join text support

ykuzmenko73:


ykuzmenko73:
[Reader with selection support does not join words with same style](https://trello.com/c/V5Spps48/921-reader-with-selection-support-does-not-join-words-with-same-style)

PR #1215, opened 2023-06-08
flowc: nim: some fixes of unicode handling

dmitry-vlasov:
Fix causes passing of test:

// cpp, java, js: [55297]
println(s2a(fromCharCode(55297)));

// cpp, java, js: [0]
println(string2utf8(fromCharCode(0)));


sfandersen (approved):
Looks good.

PR #1216, opened 2023-06-09
Flowc nim struct casts

dmitry-vlasov:
Polymorphic structs casts are extracted to separate functions to enable recursive casts like:
`cast(t: Tree<int, FiFunctionDec> -> Tree<int, FiDeclaration>)`


alstrup (approved):

sfandersen:
I have no comments.

PR #1221, opened 2023-06-15
Flowc cpp3 refcounter 2

dmitry-vlasov:


alstrup (approved):
Very big change of course, but if Java is tested and works like before, LGTM

PR #1233, opened 2023-07-27
Change z-index for background image.

AlexandrRzz:


www/flowjspixi.css
```diff
@@ -242,6 +242,6 @@ input.no_numeric_arrows::-webkit-outer-spin-button {
 	z-index: -2;
 }
 
-.textBackgroundWidget {
+.backgroundImage {
```
NikitaFil, line 245: It would rather be better 'textBackgroundLayer'

NikitaFil (approved):

PR #1241, opened 2023-08-18
strReplace with case insensitive search of a substring

pzotov:
task https://trello.com/c/G9iuNI4x

### DIFF
--- a/lib/string.flow
+++ b/lib/string.flow
@@ -487,28 +487,28 @@ strCountOfHelper(text : string, substr : string, index : int, count : int) -> in
 native strReplace : (string, string, string) -> string = Native.strReplace;
 strReplace(s : string, old : string, new : string) -> string {
 	if (old == "") s
-	else if (strlen(s) > 500) list2string(doStrReplace(s, 0, old, new, makeList()))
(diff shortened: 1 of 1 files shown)


lib/string.flow
```diff
@@ -512,6 +514,35 @@ doStrReplace2(prefix : string, suffix : string, old : string, new : string) -> s
...
+		hit = strRangeIndexOf(toLowerCase(s), old, i, n);
+		if (hit >= 0) {
+			before = substring(s, i, hit - i);
+			nacc = Cons(new, Cons(before, acc));
+			doStrIReplace(s, hit + strlen(old), old, new, nacc)
+		} else Cons(strRight(s, i), acc);
+	} else acc;
+}
+
+doStrIReplace2(prefix : string, suffix : string, old : string, new : string) -> string {
```
yegordovganich, line 535: intead of adding new it's better to extend current doStrReplace2 with passing there fn like toLowerCase or idfn. the same for doStrIReplace.
And also check if there are native sibling at haxe StringTools

yegordovganich (approved):

PR #1242, opened 2023-08-18
Added functions for timezone ids.

sergeymaksimov:


lib/timezones.flow
```diff
@@ -1,55 +1,181 @@
-import ds/array;
+import algorithms;
 
 export {
 	Timezone(
-		title : string,
-		offset : double,
-		description : string,
+		id : string,
```
sfandersen, line 5: It might be nice with a TimezoneId struct, so it becomes: 

Timezone(
   id: TimezoneId,
   name: string
);

TimezoneId(id: string);

This way functions that uses timezones will take an TimezoneId, and not just a string.

sfandersen (approved):

PR #1245, opened 2023-08-30
Flowc lsp limit flowc instances

dmitry-vlasov:


Eugeny48 (approved):
I do not know why we need this, I have never seen such a problem that flowc_lsp starts multiple flowc instances. Still this code looks good for me.

PR #1252, opened 2023-09-22
strSplit refactoring

yegordovganich:
- make comments clearer;
- get rid of doStrSplit2;
- fix strSplitLeave case with empty separator;

without any change in functionality


pzotov (approved):
Looks good

PR #1259, opened 2023-10-23
Tropic. Fixed dynamic words order in DOM

AlexandrRzz:
https://trello.com/c/2bBn5Kzm/19735-wrong-words-order-in-wigi-paragraph-tropic-mode

### DIFF
--- a/lib/tropic/tropic_paragraph.flow
+++ b/lib/tropic/tropic_paragraph.flow
@@ -600,8 +600,6 @@ positionInteractiveWords(
 	)
 }
 
-wordGroupId = ref 0;
-
 // Groups words by id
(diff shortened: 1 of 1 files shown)


lib/tropic/tropic_paragraph.flow
```diff
@@ -584,6 +600,8 @@ positionInteractiveWords(
 	)
 }
 
+wordGroupId = ref 0;
```
ykuzmenko73, line 603: make it's name more specific, like tropicParagraphWordGroupId and add comment why it is done using global ref

ykuzmenko73 (commented):
LGTM

PR #1264, opened 2023-11-07
Fix horizontal scroll gap

ahrytsavets:
https://trello.com/c/3sHvxYMu/23265-scroll-bar-overlapping-items-in-profile-settings

### DIFF
--- a/lib/tropic/tscroll.flow
+++ b/lib/tropic/tscroll.flow
@@ -615,10 +615,9 @@ TScroll(manager : TManager, content : Tropic, box : Tropic, style : [TScrollStyl
 						xMoving,
 						yMoving,
 						scrollBoxHover,
-						fif(const(overlapContent), zero, fwidth(scrollbarGaps)),
+						if (overlapContent || !isRTL) zero else fwidth(scrollbarGaps),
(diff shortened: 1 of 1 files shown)


lib/tropic/tscroll.flow
```diff
@@ -615,6 +615,7 @@ TScroll(manager : TManager, content : Tropic, box : Tropic, style : [TScrollStyl
 						xMoving,
 						yMoving,
 						scrollBoxHover,
+						fif(const(overlapContent), zero, fwidth(scrollbarGaps)),
```
NikitaFil, line 618: it's just `if`

lib/tropic/tscroll.flow
```diff
@@ -1294,6 +1296,7 @@ createTScrollBars(
 					)
 				)
 				|> (\t -> TFixSize(t, makeScrollbarBox(boxSize)))
+				|> (\t -> TBorderTDir(zero, zero, if (horizontal) xScrollGap else zero, zero, t, isRTL))
```
NikitaFil, line 1299: It would be better to move the vertical scrollbar to the left here instead of moving the whole group in #621 and compensate for horizontal here.

PR #1269, opened 2023-11-16
Prevent clicking on image if it's scrolling

AlexandrRzz:
https://trello.com/c/fTFeHeVM/19718-mobile-view-fix-swiping-in-forums

### DIFF
--- a/lib/formats/html/html2form.flow
+++ b/lib/formats/html/html2form.flow
@@ -608,12 +608,13 @@ getHtmlPictureMagnifyWithCloseButton(
 		);
 
 		cursorB = make(DefaultCursor());
+		isNotScrollB = make(true);
 		form =
 			Interactive(
 				[
-					MouseClick(nop1),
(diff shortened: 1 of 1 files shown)


lib/formats/html/html2form.flow
```diff
@@ -608,12 +608,13 @@ getHtmlPictureMagnifyWithCloseButton(
 		);
 
 		cursorB = make(DefaultCursor());
+		isNotScrollB = make(true);
 		form =
 			Interactive(
 				[
-					MouseClick(nop1),
+					MouseClick(\__ -> nextDistinct(isNotScrollB, true)),
```
NikitaFil, line 615: MouseClick is called on 'mouseup', so it shouldn't help. MouseDown would be better.

NikitaFil (approved):

lib/formats/html/html2form.flow
NikitaFil: we realized that MouseClick and MouseDown are the same

PR #1276, opened 2023-11-23
Flowc js array bounds check

dmitry-vlasov:


Eugeny48 (approved):
It looks good, but maybe "debug=1" should enable "js-array-bounds=1" flag?

alstrup:
Agree that this should also be default when using "debug=1"

dmitry-vlasov:
`js-array-bounds=1` option is replaced by `debug=1`, `js-array-bounds` option is removed

PR #1278, opened 2023-11-28
Add MArrowNavigationArray function

SLemishko:
https://trello.com/c/ZKpBQYEa/23317-issues-in-mobile-tablet-match-and-categorize-probes


lib/material/internal/material_util.flow
```diff
@@ -1117,7 +1121,27 @@ mColorLerp(previous : MColor, current : MColor, percent : double) -> MColor {
 }
 
 // Navigate through list by arrows, not by tab. Wrap each item of list into this.
```
NikitaFil, line 1123: Move the comment back to MArrowNavigationItem and add the proper one to this function

PR #1285, opened 2023-12-07
Wigi: Fix for detached spaces

ahrytsavets:
https://trello.com/c/rZR5GXHL/949-the-readers-annotated-by-gpt-has-some-unwanted-highlighting-at-the-start-of-a-paragraph-which-cannot-be-deleted-by-deleting-the


lib/tropic/tropic_paragraph.flow
```diff
@@ -885,7 +885,7 @@ commonWordConstructor(
 }
 
 inspectWordView(tr : Tropic, inspector : ParaElementInspector, letterSpacing : double) -> Tropic {
-	if (isConstantSizeTropic(tr)) {
+	(if (isConstantSizeTropic(tr)) {
```
ykuzmenko73, line 888: discuss with @NikitaFil if there is better solution that TShow
otherwise LGTM

PR #1288, opened 2023-12-08
Implemented *ByFirst/*BySecond helpers family

NikitaFil:
https://trello.com/c/kOC8m9ed/816-implement-byfirst-bysecond-helpers-family

See test_first_second_helpers.flow for testcase

### DIFF
--- a/lib/material/material_gui.flow
+++ b/lib/material/material_gui.flow
@@ -1,4 +1,5 @@
 import material/material;
+import material/internal/tropic_size_util;
 
 export {
 	// Bunch of helpers to inspect sizes. Consider using MInspect also.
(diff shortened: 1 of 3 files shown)


lib/material/material_gui.flow
```diff
@@ -454,14 +461,103 @@ MAttachHeight2(m : Material, height : DynamicBehaviour<double>, c : Material) ->
...
 MAttachToFirst(first : Material, second : Material) -> Material {
-	wh = make(WidthHeight(0.0, 0.0));
+	wh = makeWH();
 	MGroup2(
 		MAttachBox(first, wh),
-		MAvailable(
-			MFixSize(second, MFixed(0.0, 0.0)),
-			TSized(wh)
-		)
+		MAvailable(second, TSized(wh))
```
alstrup, line 467: If the second is bigger than the first, and might be growing and changing over time, will that not cause the available for the first to change, against the desire? The idea is that the first one will NOT be affected by the second in any case.

PR #1293, opened 2024-01-22
Fixed dimmer in images with custom size less then actual size.

sergeymaksimov:
Before:
Screenshot_20240122_124858
After:
Screenshot_20240122_125839


NikitaFil (approved):
LGTM

PR #1295, opened 2024-01-22
Flowc cpp3 inline fold

dmitry-vlasov:


tools/flowc/statements/fs_inline.flow
```diff
@@ -0,0 +1,599 @@
...
+		default: s;
+	}
+	switch (s2.next) {
+		Some(nx): fsSetNext(s2, fsInlineStatsDo(nx, worth_inlining, closure_lets, inlined_names, new_id, debug));
+		None(): s2;
+	}
+}
+
+fs_inlined_operators = pairs2tree([
+	Pair("fold", FsInlineOperator(
```
alstrup, line 162: Can we define some listf of names in flow.config to define what functions to inline? It seems we can automatically detect the lambda args from the types?

alstrup (approved):
Generally approved, but as discussed in chat, I would like to let the user configure what functions to inline in flow.config or a command line option.

PR #1303, opened 2024-02-02
Implement set/getJsonNestedObjectOrArrayField

stanislavpodolia:
This PR adds set/getJsonNestedObjectOrArrayField to allow editing nested Json fields including JsonArray indexes.
Extending setJsonNestedField with JsonArray indexes can break existing code because fields like "0" are treated as JsonObject members and new logic will set them as JsonArray values, there are different cases when you need one or another behavior, that's why new function is introduced.
Related PRs: https://git.area9lyceum.com/Lyceum/lyceum/pulls/7814

### DIFF
--- a/lib/formats/json/json.flow
+++ b/lib/formats/json/json.flow
@@ -97,6 +97,14 @@ export {
 	// The setter to the nested field, dual to getJsonNestedField
 	setJsonNestedField(json: Json, fields: [string], value : Json) -> Json;
 
+	// Access to a nested field like obj.f1.f2.f3, where fields = [f1, f2, f3] including JsonArray indexes
+	getJsonNestedObjectOrArrayField(json: Json, fields: [string]) -> Json;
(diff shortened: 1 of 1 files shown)


lib/formats/json/json.flow
```diff
@@ -97,6 +97,14 @@ export {
 	// The setter to the nested field, dual to getJsonNestedField
 	setJsonNestedField(json: Json, fields: [string], value : Json) -> Json;
 
+	// Access to a nested field like obj.f1.f2.f3, where fields = [f1, f2, f3] including JsonArray indexes
+	getJsonNestedObjectOrArrayField(json: Json, fields: [string]) -> Json;
```
shkel, line 101: The Json standard allows numbers to be used as object field names. Which makes the expected function result not entirely clear.
I would move this function to the app repository and remove the unused one (setJsonNestedObjectOrArrayField).

PR #1304, opened 2024-02-02
Flowc cpp3 dynamic funcs

dmitry-vlasov:


platforms/java/com/area9innovation/flow/Native.java
```diff
@@ -1137,7 +1137,29 @@ public static final boolean isStructName(String name) {
 		return FlowRuntime.struct_ids.containsKey(name);
 	}
 
-	public static final int extractFuncArity(Object fn) {
```
alstrup, line 1140: Why is it a good idea to have this magic with strings?

platforms/java/com/area9innovation/flow/Native.java
```diff
@@ -2269,7 +2291,7 @@ public static final String urlDecode(String s) {
 	@SuppressWarnings (value="unchecked")
 	public static final Object hostCall(String name, Object[] args) {
 		if (!host_call_funcs.containsKey(name)) {
-			addHostCall(name, args);
+			addHostCall(name, args.length);
```
alstrup, line 2294: Why this?

alstrup (commented):
All the Java changes look risky to me - what is the point of changing all of that?
I would probably prefer a separate PR for this, unless you can explain it and 
@Eugeny48 agrees.

Eugeny48 (commented):
I'm not sure why we need all these dynamic functions in Java but they look good to me.

platforms/java/com/area9innovation/flow/Native.java
dmitry-vlasov: The main purpose of this PR is to prepare runtime for dynamically calling functions by their name as string. It may be used in `eval` function. Now `eval` interprets all calls in AST down to natives, which are called, but with calls by name as string enabled (i.e. with `hostCall`), `eval` may use many of functions from runtime directly.

platforms/java/com/area9innovation/flow/Native.java
dmitry-vlasov: Because `addHostCall` uses only length of array of arguments, so there's no need to pass arguments as values.

alstrup:
But why is this mixed up with c++ changes? It seems to be two separate things. Is there really a need to have runtime call of functions in Java?

dmitry-vlasov:
> But why is this mixed up with c++ changes? It seems to be two separate things. Is there really a need to have runtime call of functions in Java?

I enable dynamic functions calling both in java and c++ backend as well. Tests on these features pass on both backends.

tools/flowc/backends/cpp3/fi2cpp3_compile.flow
```diff
@@ -22,16 +22,41 @@ fiModule2cpp3(module : FsModule, gctx0: Cpp3GlobalContext) -> Cpp3CompiledModule
...
 	);
 	is_main = exists(module.functions, \fn -> fn.name == "main");
+	init_body = removeEmptyLines(concatStrings(concat(
+		map(compiled_structs, \s -> strIndent(s.init) + "\n"),
+		map(compiled_decls, \d -> strIndent(d.init) + "\n"),
+	)));
+	term_body = removeEmptyLines(concatStrings(concat(
+		map(compiled_structs, \s -> strIndent(s.term) + "\n"),
+		map(compiled_decls, \d -> strIndent(d.term) + "\n"),
+	)));
```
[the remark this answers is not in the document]
dmitry-vlasov: This place is reworked with `filtermap`. For structs you are right. `s.init` can't be empty, but for some declaratins (namely some generated helper declarations) `d.init` and `d.term` may be empty.

dmitry-vlasov:
> But why is this mixed up with c++ changes? It seems to be two separate things. Is there really a need to have runtime call of functions in Java?

I reworked `eval` function (precisely it is `fcReplEval`) so that it can utilize the runtime function from compiler, not interpret it down to the natives (as it was before). In synthetic tests it shows a significant difference in performance. Now all dynamic tests pass, interpreter works both in java and in c++ backends.

PR #1312, opened 2024-02-21
Flowc compile time consts

dmitry-vlasov:


lib/compile_time.flow
```diff
@@ -0,0 +1,20 @@
+export {
+	// Speculative function, used to make output of computed constant values during compilation.
+	// Normally these function is eliminated from program before passing it to backends.
+	outputCt(msg: string, v: ?) -> ? {
```
alstrup, line 4: call this "compileTime".

lib/compile_time.flow
```diff
@@ -0,0 +1,20 @@
...
+		v
+	}
+
+	// Speculative function, used to make forbid a compile-time computation of an exression.
+	// Normally these function is eliminated from program before passing it to backends.
+	forbidCt(v: ?) -> ? {
```
alstrup, line 10: I don't see what the point of this one is.

lib/compile_time.flow
```diff
@@ -0,0 +1,20 @@
...
+		v
+	}
+
+	// Speculative function, used to make force a compile-time computation of an exression.
+	// In case a compile-time computation  fails, the error will be reported
+	// Normally these function is eliminated from program before passing it to backends.
+	assertCt(v: ?) -> ? {
```
alstrup, line 17: This will invite people to just increase compile times in general, so are we sure we want this? At least, we have to have a way to control this.

lib/dynamic.flow
```diff
@@ -3,8 +3,9 @@ export {
 	native isArray : (value : flow) -> bool = Native.isArray;
 	// check if these are structs of the same type; isSameStructType(a,a) == isStruct
 	native isSameStructType : (value1 : ?, value2 : ??) -> bool = Native.isSameStructType;
-	// compare numbers and strings by value, and structs, arrays or closures by pointer
-	native isSameObj : (value1 : ?, value2 : ?) -> bool = Native.isSameObj;
+	// compare numbers and strings by value, and structs, arrays or closures by pointer.
+	// This native is marked io, as it behaves differntly in runtime/compile-time evaluation.
```
alstrup, line 7: Explain how it is different?

tools/flowc/const_eval.flow
```diff
@@ -0,0 +1,861 @@
...
+					a2 = cast(args[1]: FiExp -> FiDouble);
+					Some(FiDouble(a1.d * a2.d, a1.start));
+				}
+				default: None();
+			}
+		}
+		FcDivPrim(): {
+			switch (args[0]) {
+				FiInt(__,__): {
+					a1 = cast(args[0]: FiExp -> FiInt);
```
alstrup, line 638: If you bound args[0] to a name, the cast is not necessary

tools/flowc/manipulation/common.flow
```diff
@@ -13,16 +13,16 @@ export {
 	);
 
 	// General transformation of a program: on a module level
-	fiMapProgramModule(prog: FiProgram, fn: (FiModule, FiProgram) -> FiModule) -> FiProgram;
+	fiMapProgramModule(prog: FiProgram, fn: (FiModule, FiProgram) -> FiModule, concurrnetly: bool) -> FiProgram;
```
alstrup, line 16: Misspelled - also below.

tools/flowc/manipulation/optimizer.flow
```diff
@@ -102,6 +106,240 @@ makeOptEnv(config : Tree<string, string>, overrides : Tree<string, FiExp>) -> Op
...
+				FiUnsafe(__, fb,__,__): {
+					FiUnsafe(x with fallback = rec(fb));
+				}
+				FiConst(): x;
+			}
+		}
+	}
+}
+
+
```
alstrup, line 322: Excessive newlines

tools/flowc/tests/cpp3/test_compile_time.flow
```diff
@@ -0,0 +1,231 @@
+import text/binary;
```
alstrup, line 1: Is this test only relevant for Cpp, or is it a more general test case?

alstrup (changes requested):
Rename outputCt to "compileTime". We also need some way to profile in time and output size to help manage that this does not go nuts and bloat the compile time and final binaries.

lib/compile_time.flow
Eugeny48: We need this for example for JAVA: when some generated array is more than 64K, we have to split it into parts, we do not want the compiler to concatenate these arrays back.

tools/flowc/tests/cpp3/test_compile_time.flow
dmitry-vlasov: Both, c++ and java backends pass tests with optimization.

tools/flowc/const_eval.flow
dmitry-vlasov: Fixed.

tools/flowc/manipulation/common.flow
dmitry-vlasov: Fixed.

lib/compile_time.flow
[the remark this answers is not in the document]
dmitry-vlasov: `assertCt` renamed to `forceCompileTimeEval`

tools/flowc/backends/java/fi2java_compile.flow
```diff
@@ -472,7 +472,30 @@ generateJavaStructDefs(gctx : JavaGlobalContext, output_dir : string) -> void {
...
+	native_name = p.second;
+	native_full = native_host + "." + native_name;
+
+	reslist = ref makeList();
+	rlistPush(reslist, indent + "// The wrapper to a native: " + native_full + "\n");
+	rlistPush(reslist, indent + prefix + "(" + superglue(arg_locals, \loc -> loc.type + " " + loc.id, ", ") + ") {\n");
+	rlistPush(reslist, indent + "\treturn " + native_full + "(" + superglue(arg_locals, \loc -> loc.id, ", ") + ");\n");
+	rlistPush(reslist, indent + "}\n");
+
+	list2string(^reslist);
```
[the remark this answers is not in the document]
dmitry-vlasov: Using `concatStrings` now, fixed.

tools/flowc/backends/java/fi2java_compile.flow
```diff
@@ -472,7 +472,30 @@ generateJavaStructDefs(gctx : JavaGlobalContext, output_dir : string) -> void {
 	});
 }
 
-emitJavaFunction(gctx : JavaGlobalContext, csm : JavaModule, global : string, name : string, tree : FiLambda, parent : Maybe<JavaScopeContext>, indent0 : string, type : FiTypeFunction) -> string {
+emitJavaNativeWrapper(gctx : JavaGlobalContext, csm : JavaModule, global : string, name : string, p : Pair<string, string>, indent : string, type : FiTypeFunction) -> string {
```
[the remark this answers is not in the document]
dmitry-vlasov: Done.

tools/flowc/manipulation/optimizer.flow
dmitry-vlasov: Fixed.

lib/compile_time.flow
alstrup: What is the difference between outputCompileTimeEval and forceCompiletimeEval?

Also, "output" makes no sense at all. "compileTime" makes a lot of sense, like I wrote. No need to say "eval" - that is not required.

dmitry-vlasov: 1) outputCompileTimeEval: _prints_ the computed expression on the screen during compilation. If computation of an expression fails (i.e. it's not a compile-time constant), the original expression is printed. Otherwise the obtained constant is printed as an expression. The main reason for this function: make it visible for a programmer, what is being computed during compile-time. I think it's a useful function, since I would like to have a simple way of knowing values of some compile-time constants

2) forceCompiletimeEval: the directive to the compile-time evaluation engine to raise an error, if the computation of a constant during compile-time fails. On my point, this function make sense when a programmer intentionally wants to perform some computations during compile-time and wants to be sure, that these computations really succeeded. 

About 'Eval' suffix - ok, may be easily dropped.

alstrup: Isn't outputCompiletime just "compileTime(println(whatever))"? Seems completely unnecessary.

"compileTime" for the other one seems like the most logical name.

dmitry-vlasov: > Isn't outputCompiletime just "compileTime(println(whatever))"? Seems completely unnecessary.
> 
> "compileTime" for the other one seems like the most logical name.

Currently `compileTime(println(whatever))` won't work, because `println` is IO native (not pure), i.e. is not computed at runtime. Also `println` doesn't return value, thus can't be used as an intermediate layer, which prints a value and then passes it further. 

Another reason, why `compileTime` doesn't replace `outputCompileTime` is that currently the model of compile-time computations is implicit, i.e. the compiler tries to compute _any_ expression, and if succeeds, it may replace the expression with a constant. Thus, `compileTime` is implicitly put on _every_ expression, and in order to distinguish in which expression we are interested in during the compile time we need some markup function (which outputCompileTime exactly is).

dmitry-vlasov:
> Rename outputCt to "compileTime". We also need some way to profile in time and output size to help manage that this does not go nuts and bloat the compile time and final binaries.

`outputCt` is turned into `messageConstEval`, while `compileTime` is added as a separate function, which uses full power of flow interpreter to compute the value of an expression. Thus, two different modes of compile-time evaluation are established:
1) manual with `compileTime` and `compileTimeDef` (in case of compile-time failure of constant computation)
2) automatic, which tries to compute compile-time constants automatically and is invoked with `const-eval=1` option at compiler invocation. By default this feature is off.

When using `compileTime` there's no limitation on immutability or purity of functions, so a used can put `prinltn` of a value of interest into the function, which computes the value. Thus he can see the obtained value and decide if it's worth computing at compile time.

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
+export {
+	// Normally all these functions are eliminated from program before passing it to backends.
```
alstrup, line 2: What does this sentence mean?

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+
+	// Speculative function, used to ask to compute a constant value during compilation.
```
alstrup, line 4: Why is this speculative? It should request a compile time evaluation, and if fail, the compilatino should stop. That is not speculative. Just write that this causes the code to evaluate at compile time, and check that the result is a constant value. What is considered a constant value?

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+	// In case a compile-time computation fails, the error will be reported
+	compileTime(v: ?) -> ? {
+		v
+	}
+
+	// Speculative function, used to ask to compute a constant value during compilation.
```
alstrup, line 10: When would we ever need this? A program is fixed. Either it works, or it does not.

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+	// In case a compile-time computation fails, the default value is returned
+	compileTimeDef(v: ?, def: ?) -> ? {
+		def
+	}
+
+	// Speculative function, used to make output of computed constant values during compilation.
```
alstrup, line 16: Can't you just use println in the compile time part? I do not understand the need for this.

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+	messageConstEval(msg: string, v: ?) -> ? {
+		v
+	}
+
+	// Speculative function, used to print a message during compilation.
```
alstrup, line 21: Just do compileTime(println("Foo")). No need for this.

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+	printlnConstEval(msg: string) -> void {
+	}
+
+	// Speculative function, used to force a run-time computation of an exression.
+	forbidConstEval(v: ?) -> ? {
```
alstrup, line 26: What is the use of this? Do not call it const and compile time mixed. Be consistent, and call it forbidCompiletime.

lib/compile_time.flow
```diff
@@ -0,0 +1,34 @@
...
+		v
+	}
+
+	// Speculative function, used to force a compile-time computation of an exression.
+	forceConstEval(v: ?) -> ? {
```
alstrup, line 31: What is the difference from this to compileTime? I do not understand.

tools/flowc/tests/cpp3/test16.flow
```diff
@@ -26,27 +27,61 @@ Un ::= A, B;
...
+testCompare(sign: string, structs: [TestStruct], op: (TestStruct, TestStruct) -> bool) {
+	iter(structs, \s1-> iter(structs, \s2 -> 
+		println(toString(s1) + sign + toString(s2) + ": " + b2s(op(s1, s2)))
+	));
+}
+
+printlnAndPass(msg: string, v: ?) -> ? {
+	println(msg + ": " + toString(v));
+	v
+}
```
Eugeny48, line 39: FYI, we have `pprint` function https://github.com/area9innovation/flow9/blob/c85c358e81e13d3f37d4c73be8b0164770a50ac7/lib/runtime.flow#L477

dmitry-vlasov: Thanks, printlnAndPass is replaced with pprint in tests

PR #1316, opened 2024-03-07
JS perfomance tune: compareByValue

ykuzmenko73:
https://trello.com/c/UprvRDzc/23905-implement-nested-json-string-fjson-compression

https://github.com/area9innovation/innovation/pull/3751

we are using json2fjson which heavely uses lookupTree with Json as key. it gives huge amount of calls to compareByValue with objects.
this PR gives significant performance improvement, for example:
was: 10300 ms
tuned compareByValue: 8500 ms
most time spent in compareByValue (about 6640 ms, 1100 in which - Reflect.field), it is possible to speed it up, because for example a lot of time is spent in Object.prototype.hasOwnProperty.call which is not required because we can check it against undefined, it saves ~15% of total time and up to 30-40% of compareByValue’s time


NikitaFil (approved):
LGTM

PR #1317, opened 2024-03-11
RenderSupport: Add preventStuckModifierKeys to onpointerdown

stanislavpodolia:
In some cases modifier keys get stuck when browser loses focus but "blur" event is not fired, in particular when "Mission control" is triggered on mac. This pr adds preventStuckModifierKeys to onpointerdown to check stuck modifiers to fix the issue in this card: https://trello.com/c/HgWoJyV2/410-multiselection-is-often-the-default-on-mac


NikitaFil (approved):
LGTM, but it looked erroneously and not so obvious why we apply parseKeyEvent for PointerEvent. Maybe worth leaving a comment for the future.

PR #1320, opened 2024-03-19
QT + Java: Implement getApplicationArguments + build macOs binary

vereschagai:


platforms/js/Native.hx
```diff
@@ -361,6 +361,10 @@ class Native {
 		return "";
 	}
 
+	public static function getApplicationArguments() : Array<String> {
+		return new Array();
```
alstrup, line 365: For node, we could do something, but not required for now.

alstrup (approved):

sfandersen:
It works with Java, but with flowcpp I get this error when I try to convert the arguments to JSON. 

IInvalid Tag: arr (array expected, void found)
Error during last opcode [0x0]: 7
DataStack.size() = 521, CallStack.size() = 0hp = ffffbc4, HeapStart = 0x3a64, HeapEnd = 0x10003a64
#0  0x000303ff in testArgs$1
#1  0x00030323 in testArgs
#2  0x00037010 in main
#3  0x0003d33b in --end--$init
#4  0x00000084 in $entry_thunk
Run with --no-jit to get line numbers in this callstack

It can be reproduced with this test: 
https://github.com/area9innovation/innovation/blob/master/lib/cmd_args/cmd_args_test.sh

PR #1322, opened 2024-03-27
Implement minAFn/maxAFn helpers

ahrytsavets:


yegordovganich (commented):
good, but you can optimize it even more with comparator fn, where it's min or max

PR #1326, opened 2024-04-15
tropic_util : minor refactoring

shkel:
the same filtermap , but without wrappers. @NikitaFil what do you think? Is code cleanliness optimization worth it?
image


NikitaFil (approved):
Do I understand correct, with filtermap implementation so many garbage was generated, that GC delays overshadow time, saved by mapping array once instead of twice? If so and if it reproduces constantly - then agree.

PR #1327, opened 2024-04-15
Flowc cpp3 advances 2

dmitry-vlasov:


lib/ds/vector.flow
```diff
@@ -195,7 +218,7 @@ array2vector(arr : [?]) -> Vector<?> {
 
 // Set of flow fallbacks for natives:
 
-makeNativeVector(capacity : int) -> native {
+makeNativeVector(capacity : int, __: [?]) -> native {
```
Eugeny48, line 221: Could you fix the compiler instead of using unnecessary parameters?

PR #1328, opened 2024-04-16
TextClip: Ignore transform if useCheck is true

stanislavpodolia:
This PR skips applying transform to `textNodeWidth` in `updateTextWidth` if `useCheck=true`. `getTextNodeMetrics` with `useCheck=true` calls `updateTextNodesWidth` which attaches the svg to the body in `updateTextNodeWidth` ignoring transform applied to the text clip. As result we should skip dividing width by this transform as it was never applied. Not sure if this is a correct and only place to ignore transform.
Added a test `lib/material/tests/test_text_metrics_with_scale.flow`
without fix:
image
with fix:
image

### DIFF
--- a/lib/material/tests/test_text_metrics_with_scale.flow
+++ b/lib/material/tests/test_text_metrics_with_scale.flow
@@ -1,0 +1,8 @@
+import material/material_ui;
+
+main() {
+	setRendererType("html");
+	manager = makeMaterialManager([]);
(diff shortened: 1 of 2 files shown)


NikitaFil (approved):
Thanks. Indeed, we shouldn't apply transform in useCheck=true mode. I reworked fix to keep width calculation of old mode together.

PR #1337, opened 2024-05-08
TextClip: Disable useCheck if escapeHTML is true

stanislavpodolia:
This PR disables useCheck for TextClip in case escapeHTML is true. Attached a test to reproduce the problem:
`lib/material/tests/escape_html_test.flow`
Before fix:
image
After fix:
image

### DIFF
--- a/lib/material/tests/escape_html_test.flow
+++ b/lib/material/tests/escape_html_test.flow
@@ -1,0 +1,9 @@
+import material/material_ui;
+
+main() {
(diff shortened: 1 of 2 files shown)


NikitaFil (commented):
1) I guess you meant if escapeHTML is false in the title
2) Width is fine if we give font a time to load
3) I'm fine with height update disabling but let's do it in another way. I might fixed it later and enable again.

PR #1339, opened 2024-05-17
Introduce Math Formula external block

AlexandrRzz:


platforms/js/MathJaxClip.hx
```diff
@@ -0,0 +1,56 @@
+import js.html.URL;
```
NikitaFil, line 1: Review imports, some of them may be redundant

NikitaFil (approved):
Generally, looks good

PR #1346, opened 2024-06-17
[splitByNumber] Use tail recursion to avoid out of memory error

ClaudioChiarella:
During the test with Bridge's fixup for creator, we see that browser crash if there are 2M of contents ids that should be slitted in group of 1k ids.
The cause is the recursion in splitByNumber, so I modify it to use tail recursion

### DIFF
--- a/lib/ds/array.flow
+++ b/lib/ds/array.flow
@@ -614,16 +614,18 @@ splitByNumber(a : [?], n : int) -> [[?]]{
 	if (length(a) <= n || n <= 0) {
 		[a];
 	} else {
-		// Use tail recursion to avoid memory leaks (e.g. with 2M elements in a splitted by 1k)
-		splitByNumber1(subrange(a, n, length(a)-n), n, [subrange(a, 0, n)]);
+		// Use tail recursion to avoid out of memory error (e.g. with 2M elements in a splitted by 1k)
+		splitByNumber1(a, 0, n, []);
 	}
 }
 
-splitByNumber1(a : [?], n : int, acc : [[?]]) -> [[?]]{
+splitByNumber1(a : [?], i : int, n : int, acc : [[?]]) -> [[?]]{
 	if (length(a) <= n || n <= 0) { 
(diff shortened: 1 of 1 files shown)


andrejefimov (changes requested):
It is not a memory leak, but an out of memory. Also it would be more efficient not to copy the tail on each iteration, but just pass the whole array, a starting index and a step

splitByNumber1(a : [?], i: int, n : int, acc : [[?]]) -> [[?]] {
.... 
    splitByNumber1(a, i+n, n, concat(acc, [subrange(a, i, n)]))

lib/ds/array.flow
```diff
@@ -611,11 +611,21 @@ split(a : [?], f : (?) -> bool) {
...
 	} else {
-		concat([subrange(a, 0, n)], splitByNumber(subrange(a, n, l-n), n))
+		// Use tail recursion to avoid out of memory error (e.g. with 2M elements in a splitted by 1k)
+		splitByNumber1(a, 0, n, []);
+	}
+}
+
+splitByNumber1(a : [?], i : int, n : int, acc : [[?]]) -> [[?]]{
+	if (length(a) <= n || n <= 0) { 
+		concat(acc, [a]);
```
andrejefimov, line 624: I'd drop this: you've already checked it in `splitByNumber`, and `splitByNumber1` should never be used directly

lib/ds/array.flow
```diff
@@ -611,11 +611,21 @@ split(a : [?], f : (?) -> bool) {
...
+	} else if (length(a) <= i) {
+		acc;
+	} else {
+		splitByNumber1(a, i + n, n, concat(acc, [subrange(a, i, n)]))
```
andrejefimov, line 628: `concat(acc, [subrange(a, i, n)])` can be replaced by `arrayPush(acc, subrange(a, i, n))`

andrejefimov (approved):
Looks good, but a couple of minor things to consider

PR #1352, opened 2024-07-02
Flowc cpp3 add master key natives

dmitry-vlasov:
Master key c++ sources are compilable to quite far extend.
All Necessary natives stubs are added.


lib/sys/interactive.flow
```diff
@@ -200,3 +200,21 @@ handleRealEvents(l : EventHandler, clip : native, clipalive : ref bool, stage :
 	}
 }
 
+// Fallbacks for natives.
+getMouseX(clip : native) -> double {
+	println("TODO: implement 'getMouseX' native - a native fallback stub is used");
```
alstrup, line 205: Drop the " - a native fallback stub is used" just to save size in the output. The first part is sufficient for search/replace when it appears.

alstrup (approved):
Just reduce the native texts to make the binaries smaller and this is OK.

lib/sys/interactive.flow
dmitry-vlasov: Done

PR #1360, opened 2024-07-21
Flowc cpp3 implement db api

dmitry-vlasov:


Eugeny48:
MKS uses another implementation in Java target -- DatabaseSValue https://github.com/area9innovation/innovation/blob/564472011a7c920af2786b112588cf07066859db/components/master_key/server/database_svalue.flow.

PR #1366, opened 2024-08-01
Do not use `fold` in `fmaxA`, `fminA`

Eugeny48:
Do not use `fold` in `fmaxA`, `fminA` because it causes a crash when the array is longer than 10K.

But maybe we still need to use `fold` with short arrays?

hotfix for lyceum is here https://git.area9lyceum.com/Lyceum/lyceum/commit/ee1ccced0d7b33cb0d2740b75e42adbe143135f2
trello https://trello.com/c/dWCgbsTq/25448-attempting-to-load-a-big-class-over-10k-learners-crashes

### DIFF
--- a/lib/fusion.flow
+++ b/lib/fusion.flow
@@ -1843,33 +1843,6 @@ fmerge(a : [Transform<?>]) -> Transform<[?]> {
 		ConstBehaviour([])
 }
 
-fmergeBalanced(a : [Transform<?>]) -> Transform<[?]> {
-	if (length(a) > 0) {
-		foldTreeBinary(
-			pairs2tree(mapi(a, \i : int, v : Transform<?> -> Pair(i, Some(v)))),
-			\__, v : Maybe<Transform<?>>, l : Maybe<Transform<[?]>>, r : Maybe<Transform<[?]>> -> {
-				maybeMap(v, \vv : Transform<?> -> eitherFn(l,
-					\ll -> eitherMap(r,
-						\rr : Transform<[?]> -> {
-							fselect3(ll, vv, rr, \lll, vvv, rrr -> concat3(lll, [vvv], rrr))
-						},
-						fselect2(ll, vv, FLift2(arrayPush))
-					),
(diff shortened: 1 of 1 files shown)


Eugeny48:
[Attempting to load a big class (over 10k learners) crashes](https://trello.com/c/dWCgbsTq/25448-attempting-to-load-a-big-class-over-10k-learners-crashes)

Eugeny48 (commented):
LGTM

lib/fusion.flow
```diff
@@ -1839,6 +1843,33 @@ fmerge(a : [Transform<?>]) -> Transform<[?]> {
 		ConstBehaviour([])
 }
 
+fmergeBalanced(a : [Transform<?>]) -> Transform<[?]> {
```
alstrup, line 1846: I guess we do not need this one after all.

alstrup:
I like the new one even if not faster. I guess the main benefit is that we do not crash with stack limit.

lib/fusion.flow
```diff
@@ -1530,44 +1530,26 @@ fmin3(b1 : Transform<?>, b2 : Transform<?>, b3 : Transform<?>) -> Transform<?> {
...
-			},
-			None()
-		)
-		|> (\f -> either(f, const(def)))
+		fMaxMin(v, fmin)
 	else
 		ConstBehaviour(def)
 }
 
+fMaxMin(v : [Transform<?>], fn : (Transform<?>, Transform<?>) -> Transform<?>) -> Transform<?> {
```
Eugeny48, line 1545: I think it can be reworked without `subrange`, just pass left and right indexes to the function.

lib/fusion.flow
yegordovganich: removed

lib/fusion.flow
yegordovganich: done

NikitaFil (approved):
Seems now it is ready for merging

PR #1369, opened 2024-08-08
MWigiParagraph. Apply underline/link styles in case words are not joined

NikitaFil:
https://trello.com/c/eCLGjhrT/25479-underline-is-ignored-in-review-and-history


ykuzmenko73:
Will be fixed by @AlexandrRzz in separate PR

PR #1376, opened 2024-08-27
RtrimWords in lines for Right alignment

AlexandrRzz:
https://trello.com/c/S8FpPc9E/25394-prime-graphics-right-alignmented-text-are-not-presented-in-line-after-an-automatically-line-break


ahrytsavets (approved):
LGTM

PR #1379, opened 2024-08-30
MMultiAudioPlayer: start all fragments muted in Safari

pzotov:
Task https://trello.com/c/nBXxF9wN/25622-coach-sometimes-plays-twice-on-top-of-one-another-using-safari

Sometimes Safari on MacOS and iPad launches all fragments at once and they can not be paused by isPlay[i] (because it is false at that moment).
So for Safari I suggest starting all the fragments quietly, pausing them, and then returning the volume to the initial value and starting the whole playlist from the beginning if necessary (if autoplay is on)

@NikitaFil @stanislavpodolia any objections?

### DIFF
--- a/lib/material/material_ui.flow
+++ b/lib/material/material_ui.flow
@@ -676,7 +676,7 @@ MMultiAudioPlayer(filenames : [string], mStyle : [MMultiAudioPlayerStyle]) -> Ma
 				default : None()
 			}
 		);
-		isSafari = isSafariBrowser() && js;
+		isSafari = isSafariBrowser();
 
 	  	// FVideoStyle properties
 	 	lenEx = extractStruct(fvStyle, FVideoLength(make(0.))).length;
--- a/lib/material/material_ui.flow
+++ b/lib/material/material_ui.flow
@@ -677,6 +677,7 @@ MMultiAudioPlayer(filenames : [string], mStyle : [MMultiAudioPlayerStyle]) -> Ma
 			}
 		);
 		isSafari = isSafariBrowser();
+		allFragmentsInitiated = ref false;
 
 	  	// FVideoStyle properties
 	 	lenEx = extractStruct(fvStyle, FVideoLength(make(0.))).length;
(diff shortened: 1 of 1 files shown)


lib/material/material_ui.flow
```diff
@@ -676,6 +676,7 @@ MMultiAudioPlayer(filenames : [string], mStyle : [MMultiAudioPlayerStyle]) -> Ma
 				default : None()
 			}
 		);
+		isSafari = isSafariBrowser() && js;
```
NikitaFil, line 679: isSafariBrowser already checks for js.

lib/material/material_ui.flow
```diff
@@ -839,7 +861,7 @@ MMultiAudioPlayer(filenames : [string], mStyle : [MMultiAudioPlayerStyle]) -> Ma
 							}
 							changeAudioPlaying(newIndex, getValue(playEx));
 
-							[
+							if (isSafari) [] else [
```
NikitaFil, line 864: Why do we skip this subscription? To prevent changing from play to paused icon visually? Or to prevent from missing initial playEx state?

How will be connection  isPlay -> playEx work for Safari if video stops for some reason from native video side (apart from it ends) ?

pzotov: > How will be connection isPlay -> playEx work for Safari

This connection sets playEx to false when we change isPlay[i]. When we pause started fragments, we always set playEx = false, so autoplay will never work in Safari.

We change the playEx value when we have completed all fragments in the playlist here. I think this is enough https://github.com/area9innovation/flow9/blob/master/lib/material/material_ui.flow#L715

NikitaFil: Video can be stopped by some other reason, but in this case it still be in 'playing' state on our level.

We can just save playEx initial value (the same way as volume) and use it in line 841

pzotov: > We can just save playEx initial value (the same way as volume)

No, this value may change after the player has been rendered. I'll rework this fconnect

pzotov: reworked. In Safari, these values will be connected after all fragments are initiated

PR #1382, opened 2024-09-05
TRenderParagraph. Update on text width changed.

NikitaFil:
https://trello.com/c/UG7DbN5t/25300-the-text-extends-beyond-the-frame


ykuzmenko73 (commented):
LGTM but hard to predict if it will work...

PR #1385, opened 2024-09-12
Calendar. Add CalendarGetHourOnEmptyCell

DMLatrofa:


DMLatrofa:
[Pre-select time when clicking on a particular cell within the calendar week view](https://trello.com/c/AIMFeR7r/2386-pre-select-time-when-clicking-on-a-particular-cell-within-the-calendar-week-view)

lib/material/extra/calendar/calendar.flow
```diff
@@ -928,15 +929,16 @@ dayEventsPopup(
 	close = make(false);
 	dayEvents = make(dayEventsStatic);
 
-	curTime = stamp2time(timestamp());
```
sergeymaksimov, line 931: Looks like it will break current behaviour. Currently if you add new event the time would be rounded current time which we get from timestamp(). But if you use new curTime it will be current date with time 0:00:00. So default event time will always be rounded 0:00:00, instead of rounded current time.

DMLatrofa: what about using this kind of function instead of using date2time?

date2currentTime(date : Date) -> Time {
  Time(getCurrentTime() with year = date.year, month = date.month, day = date.day)
}

lib/material/extra/calendar/calendar.flow
```diff
@@ -928,17 +933,16 @@ dayEventsPopup(
...
-	dayNumber = curDate.day;
-	monthNumber = curDate.month;
-
 	roundTime = \time -> {
-		roundInterval = 30;
-		roundedMinutes = (time.min / roundInterval + 1) * roundInterval;
-		flooredTime = Time(time with min = 0, sec = 0);
-		stamp2time(time2stamp(flooredTime) + min2ms(roundedMinutes));
+		if (i2b(time.min) || i2b(time.sec)) {
```
sergeymaksimov, line 937: What if time will be with 0 min and 0 sec without CalendarGetHourOnEmptyCell() style? Not likely but could happen. Then it won't work as it should.
Maybe just check styles for CalendarGetHourOnEmptyCell() style?

DMLatrofa: I've added check for getHourOnClick

sergeymaksimov: Shouldn't it be just ```if (!getHourOnClick) {```?

lib/material/extra/calendar/calendar.flow
```diff
@@ -924,21 +929,21 @@ dayEventsPopup(
...
-	curTime = stamp2time(timestamp());
-	dayNumber = curDate.day;
-	monthNumber = curDate.month;
-
 	roundTime = \time -> {
-		roundInterval = 30;
-		roundedMinutes = (time.min / roundInterval + 1) * roundInterval;
-		flooredTime = Time(time with min = 0, sec = 0);
-		stamp2time(time2stamp(flooredTime) + min2ms(roundedMinutes));
+		if (getHourOnClick && (i2b(time.min) || i2b(time.sec))) {
```
sergeymaksimov, line 938: Shouldn't it be just ```if (!getHourOnClick) {```?

Maslyanko:
Хватит, пожалуйста! Нам это же сдавать надо

lib/material/extra/calendar/calendar.flow
DMLatrofa: hoursGridStyle = extractStruct(styles, HoursGridForWeekView(-1.0, 0, 24));
showHoursGrid = hoursGridStyle.cellSize > 0.0 && rangeType == 2;
getHourOnClick = containsStruct(styles, CalendarGetHourOnEmptyCell()) && showHoursGrid;

if we want to use only getHourOnClick i have to do more extractStruct that are not really neded

sergeymaksimov (approved):

PR #1386, opened 2024-09-13
Calendar - RTL fix

kanakin-egor:


NikitaFil (approved):
According to Material spec (https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements), such icons should indeed be reversed, so approve.

PR #1390, opened 2024-09-24
Use PreventTextNodeWidthCalculation only for RobotoItalic font

AlexandrRzz:


NikitaFil (commented):
Looks fine as hotfix, but must be synchronized with master

PR #1391, opened 2024-09-26
moves jwt to headers on from/ffrom level

MalkoVl:
Before the sending a URL to the native functions we split it to the URL without tokens and headers with tokens. It allows us to keep tokens safe and to prevent opening the urls (from the Network Dev Tool) on the new tab of the browser.

It is the most common implementation of moving jwt (and other token) to the request headers.

Before this changes we had to do this manually in every place in the code (for every media elements in the UI). Now, we don't have to worry about it, everything work automatically.

To get it works we moved some functions from other repos to the flow9. Is it OK?

And also was added new style (to the Picture and Video elements) to prevent parsing URL and leave it as is. SharedUrl().
It works for both From and FForm UIs.

Related PR in other repos:
[Innovation](https://github.com/area9innovation/innovation/pull/3991)
[Lyceum](https://git.area9lyceum.com/Lyceum/lyceum/pulls/8779)


lib/fform/fformutils.flow
```diff
@@ -2237,3 +2244,100 @@ addFAccessProperties(properties : [FAccessProperty], zorder : Transform<[int]>)
...
+
+						FPicture(url, m.size, concat(style2, [headersStyle]));
+					}
+				);
+			}
+		);
+	}
+}
+
+// FPictureApplyJwtExtraction(m : FPicture) -> FPicture {
```
yegordovganich, line 2310: make sense to remove this code?

lib/formats/uri/url_utils.flow
```diff
@@ -0,0 +1,207 @@
...
+setUnsetRequestHeadersLess() -> () -> void {
+	requestHeadersLessR := true;
+	\-> requestHeadersLessR := false;
+}
+
+isPhpUrl(url : string) -> bool {
+	strContains(url, ".php?");
+}
+
+isPhpUrlImproved(url : string) -> bool {
```
yegordovganich, line 71: I understand that you just moved this file from innovation repo, but this fn naming looks strange. It doesn't look like it really check for php

yegordovganich (approved):

NikitaFil (approved):
Generally, looks fine.

But maybe get rid of some types and not reimplement *ApplyJwtExtraction four times?

lib/fform/renderfform.flow
```diff
@@ -189,7 +189,9 @@ renderFForm(rform : FForm, zorder : Transform<[int]>) -> FRenderResult {
 				renderFParagraph(text, style);
 			}
 		}
-		FPicture(url_, wh, style): {
+		FPicture(url2, wh, style2): {
+			f = FPictureApplyJwtExtraction(FPicture(url2, wh, style2));
```
NikitaFil, line 193: Also, let's make it FPictureApplyJwtExtraction(rForm) and don't create additional object.

PR #1392, opened 2024-09-26
added cache and queue for media downloads

MalkoVl:
Optimization to load a single image/video/audio only once.


platforms/js/RequestQueue.hx
```diff
@@ -0,0 +1,73 @@
...
+
+		// Places request into the download queue
+		pendingRequests.set(url, promise);
+
+		return promise;
+	}
+
+	public function removeBlobUsage(url: String) {
+		var cnt = blobUsage.get(url);
+		if (cnt == 1) {
```
NikitaFil, line 64: >1 ?

PR #1394, opened 2024-10-02
TPdfDocumentStyle. Added TPdfWH

DMLatrofa:


DMLatrofa:
[Enable zooming into attachment directly in add treatment record view](https://trello.com/c/ZqRH96QF/2440-enable-zooming-into-attachment-directly-in-add-treatment-record-view)

vereschagai (approved):

PR #1399, opened 2024-10-08
new methods to cache size, refactoring

MalkoVl:
Due to new feature which removes tokens from the url - updated the caching of picture size


NikitaFil (approved):
LGTM

PR #1401, opened 2024-10-09
implemented validation string operations in java target.

vadlis:
In fact - copied from security.php

### DIFF
--- a/lib/password.flow
+++ b/lib/password.flow
@@ -5,6 +5,17 @@ export {
 		password : string, 
 		hash : string 
 	) -> bool = FlowPassword.validateHash;
+
+	// < 0: means no expiration
+	// 0: use default expiration time (84600 sec -> 1 day)
+	// > 0: use specified value 
+	native getPasswordValidationStringExp : (exp: int) -> string = FlowPassword.getPasswordValidationString;
+	// get validation string without expiration (create_reset_password_validation_string(null))
+	getPasswordValidationString : () -> string;
+
+	native getValidationStringExpiration : (vs : string) -> int = FlowPassword.getValidationStringExpiration;
+	native checkValidationStringExpirationExp : (vs : string, exp: int) -> bool = FlowPassword.checkValidationStringExpiration;
+	checkValidationStringExpiration : (vs : string) -> bool;
 }
 
 createPasswordHash(password : string) -> string {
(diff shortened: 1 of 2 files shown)


bvaschenko (approved):
LGTM

lib/password.flow
```diff
@@ -5,6 +5,17 @@ export {
 		password : string, 
 		hash : string 
 	) -> bool = FlowPassword.validateHash;
+
+	// < 0: means no expiration
+	// 0: use default expiration time (84600 sec -> 1 day)
+	// > 0: use specified value 
+	native getPasswordValidationStringExp : (exp: int) -> string = FlowPassword.getPasswordValidationString;
```
sfandersen, line 12: Instead of "exp" it is better to call it "expirationS". (I assume it is in seconds).

lib/password.flow
```diff
@@ -5,6 +5,17 @@ export {
...
+	// get validation string without expiration (create_reset_password_validation_string(null))
+	getPasswordValidationString : () -> string;
+
+	native getValidationStringExpiration : (vs : string) -> int = FlowPassword.getValidationStringExpiration;
+	native checkValidationStringExpirationExp : (vs : string, exp: int) -> bool = FlowPassword.checkValidationStringExpiration;
```
sfandersen, line 17: It would seem to me like the checkValidationStringExpirationExp does not need to be an native?

vadlis: agree

platforms/java/com/area9innovation/flow/FlowPassword.java
```diff
@@ -102,4 +106,68 @@ public static Boolean validateHash(String password, String hash) {
...
+
+		try {
+			JSONObject jo = new JSONObject();
+			jo.put("key", Base64.getEncoder().encodeToString(bytes));
+			if (expLocal != null) {
+				long unixTime = System.currentTimeMillis() / 1000L;
+				jo.put("exp", unixTime + expLocal);
+			};
+
+			validation = strtr(Base64.getEncoder().encodeToString(jo.toString().getBytes()), "+/=", "*_-");
```
sfandersen, line 126: I think you can just do "Base64.getUrlDecoder", that way you do not need the strtr function (If you need it, rename it to something like replaceCharaters

vadlis: I think we need this to produce the same string as security.php do. So string from java & php are the same and can be processed in both targets

sfandersen (approved):
I had a few comments. Otherwise it looks good.

platforms/java/com/area9innovation/flow/FlowPassword.java
sfandersen: Right. Agree, the * and _ are not normal base64url encode.

PR #1408, opened 2024-11-13
strRepeat : new implementation

shkel:
if cnt ~= 18000 we have max call stack error.

### DIFF
--- a/lib/string.flow
+++ b/lib/string.flow
@@ -1388,7 +1388,9 @@ toStringNoExtraQuotes(v: flow) -> string {
 }
 
 strRepeat(s : string, cnt : int) -> string {
-	if (cnt > 0) s + strRepeat(s, cnt - 1) else "";
+	// if (cnt > 0) s + strRepeat(s, cnt - 1) else "";
+
+	strGlue(generate(0, cnt, \__ -> s), "");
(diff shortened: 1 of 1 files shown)


shkel:
@alstrup  what do you think ?

alstrup (approved):
OK with me. I guess we could do some divide and concur and do bigger chunks at a time, but this works.

lib/string.flow
```diff
@@ -1388,7 +1389,49 @@ toStringNoExtraQuotes(v: flow) -> string {
...
 
 strRepeat(s : string, cnt : int) -> string {
-	if (cnt > 0) s + strRepeat(s, cnt - 1) else "";
+	if (cnt <= 0) {
+		"";
+	} else if (cnt == 1) {
+		s;
+	} else {
+		// list2string(generateList(0, cnt, \__ -> s));
+		p = generateProgressString(cnt, makeList1(s), 1, 1);
```
shkel, line 1398: @alstrup  like list2string(generateList(0, cnt, \__ -> s)) more . but maybe i didn't understand you

alstrup:
The simple one with list2string was fine. We do not need the complicated version.

PR #1430, opened 2025-01-28
Fixed a profile in the Hamburger menu gets focused and announced as a button

deltaonlyone:
https://trello.com/c/S60hmvR3/26735-profile-in-hamburger-menu-gets-focus-and-announced-as-a-button-and-list


NikitaFil (changes requested):
Will affect unrelated cases

PR #1437, opened 2025-02-04
Update around UTF8 encoding

Eugeny48:
Removed `--use_utf8_js_style` flowcpp flag, since its behavior is standard UFT8 format, use this by default;

To use outdated format introduced flowcpp flag `--use_wtf8`; Removed "js" suffix from names;

Removed parser dependency on "js_style" in cpp code: we should always create surrogate pairs for big codes when parsing UTF8 into UTF16 (internal flowcpp format), otherwise UTF16 data cannot be correctly restored back to UTF8; After removing the dependency, "base" functions (like "parse_base") was removed as they become the same as normal functions (for example "parse");

Added a simple test case; Added more tests to the bugfour application in the innovation repo;

### DIFF
--- a/platforms/common/cpp/utils/AbstractHttpSupport.cpp
+++ b/platforms/common/cpp/utils/AbstractHttpSupport.cpp
@@ -231,10 +231,8 @@ void AbstractHttpSupport::deliverResponse(int id, int status, HeadersMap headers
                 switch (rq->response_enc)
                 {
                     case ResponseEncodingUTF8:
-                        data = RUNNER->AllocateString(parseUtf8Base((const char*)pdata, count, false));
-                        break;
-                    case ResponseEncodingUTF8js:
-                        data = RUNNER->AllocateString(parseUtf8Base((const char*)pdata, count, true));
+                    case ResponseEncodingWTF8:
+                        data = RUNNER->AllocateString(parseUtf8((const char*)pdata, count));
                         break;
                     case ResponseEncodingByte:
                         data = RUNNER->AllocateString((unicode_char*)pdata, count/FLOW_CHAR_SIZE+count%FLOW_CHAR_SIZE);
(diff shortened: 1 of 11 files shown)


Eugeny48:
The PR was tested on the bugfour, before that tests failed. https://jenkins.area9innovation.com/job/App/job/Bugfour/312122/

alstrup (approved):
Great work!

platforms/common/cpp/utils/AbstractHttpSupport.cpp
```diff
@@ -737,9 +735,9 @@ StackSlot AbstractHttpSupport::deleteAppCookies(RUNNER_ARGS)
 
 ResponseEncoding AbstractHttpSupport::GetResponseEncodingFromString(std::string str)
 {
-    if (str == "utf8_js")
-        return ResponseEncodingUTF8js;
-    else if (str == "utf8") {
+    if (str == "utf8")
+        return ResponseEncodingWTF8;
+    else if (str == "wtf8") {
         return ResponseEncodingUTF8;
```
MalkoVl, line 741: Do we need to change the values ​​under these conditions (utf8 and wtf8)?

MalkoVl (commented):
Overall, it looks good for me. There are just a few questions.
Am I correct in understanding that
- utf8 with surrogate pairs is now the default utf8 encoding;
- wtf8 is backwards compatible with utf8;
- we no longer support byte/ASCII encoding (I ask only for case if somebody out of A9 uses this repo)?

Eugeny48:
@MalkoVl 
> * utf8 with surrogate pairs is now the default utf8 encoding;

Now the default is UTF8, before it was called "utf8_js" in our code.
When flowcpp makes UTF8 from a surrogate pair it will produce 1 UTF8 symbol (4 bytes).
When flowcpp makes WTF8 from a surrogate pair it will produce 3 UTF8 symbol (3+3 bytes).
Only this differs these encodes in our code.

The reverse conversion from UTF8/WTF8 to UTF16 is the same -- the code creates surrogate pairs when necessary.


> * wtf8 is backwards compatible with utf8;

No. A pair of 3 byte characters produced from a surrogate pair is invalid in standard UTF8 and should produce 2 error codes.
But in our code we allow this conversion even in UTF8.

> * we no longer support byte/ASCII encoding (I ask only for case if somebody out of A9 uses this repo)?

Nothing has changed here.

platforms/common/cpp/utils/AbstractHttpSupport.cpp
Eugeny48: Thanks. Fixed.

PR #1438, opened 2025-02-05
add ShowMDialogForFolderWithCallbacks

O-Shev:
https://trello.com/c/dbSC6YNP/26870-add-create-and-open-button-into-files-view-create-folder-menu


MalkoVl (commented):
I'm not sure this is the right place for dialogue. In the `material` folder we store the basic components of the Material, but the dialog for creating a folder is a special case. Let's do a trick. We have roughly the same dialog in the filebrowser component. We can move this dialog, for example, [here](https://github.com/area9innovation/flow9/blob/master/lib/material/internal/filebrowser/material_filebrowser_internal_view.flow) and reuse it in both, filebrowser and filesystemdb.

MalkoVl (approved):
LGTM

PR #1440, opened 2025-02-06
Readability improvements and bug fixes of csv.flow by Mindcode

andrejefimov:


alstrup:
If it has been tested, and you are confident it is correct, OK with me. It is too big for me to review in details.

alstrup (commented):
How was this tested? A trick is that you can often ask Mindcode to write a test case to illustrate the bug, and that way get some more evidence that the rework is correct.
I did not review the diff in detail, but if you are confident it is good, OK.

andrejefimov:
@alstrup I've committed some tests generated by Mindcode. Also I've run tests on real data from old cases where these issues were noticed, e.g. https://trello.com/c/9aQmv0FY/454-import-mcqs-in-csv-does-not-work

PR #1444, opened 2025-02-19
add letterSpacingSuportedSizeInspector function

SLemishko:
https://trello.com/c/cDX1vkud/27000-epub-converter-places-some-text-blocks-to-a-wrong-position-in-the-fixed-design-readers

### DIFF
--- a/lib/tropic/tropic_paragraph.flow
+++ b/lib/tropic/tropic_paragraph.flow
@@ -644,11 +644,15 @@ positionInteractiveWords(
 			viewFn = lookupTreeDef(joinedWordsConstructors, getValue(word.id), word.wordsViewConstructor);
 			wordView = switch (word.word) {
 				GeneralText(str) : {
-					TInspect(
-						if (useWigiTextDynamicMetrics) [ISize(word.inspector.size), IBaseline(word.inspector.baseline)]
-						else [],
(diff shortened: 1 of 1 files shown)


lib/tropic/tropic_paragraph.flow
```diff
@@ -944,23 +948,31 @@ commonWordConstructor(
 	);
 }
 
+letterSpacingSuportedSizeInspector(
+	inspector : ParaElementInspector,
+	letterSpacing : double,
+	view : Tropic
+) -> Tropic {
+	sizeB = make(zeroWH);
+	uns = subscribe2(sizeB, \v -> {
```
ykuzmenko73, line 957: it should be in the constuctor, not outside

ykuzmenko73 (commented):
LGTM, small comment

lib/tropic/tropic_paragraph.flow
NikitaFil: But it's you made it so in the first place. I guess, there was some reason for that.
https://github.com/area9innovation/flow9/commit/0e7407b4e288be0a359e5ca1921f7bc28fd75438#diff-88d1bfbd82d2dc7b7d301af30d100d3d2c133e53363775e8dd7de2be6be413a5R838

NikitaFil (approved):

PR #1450, opened 2025-02-27
Adjust slider width to fit captions

AlexandrRzz:
https://trello.com/c/3katnL98/20817-improve-wigi-slider-element
image


NikitaFil (approved):
LGTM. The only improvement I can see - extract duplicates to separate function.

PR #1458, opened 2025-03-19
Refactored

bvaschenko:
@NikitaFil, could be that initial intension was to apply these pipes to whole if-else code, but I'm not sure.
Could you please take a look (sorry, don't know whom else I can ask)?

### DIFF
--- a/lib/ui/buttons.flow
+++ b/lib/ui/buttons.flow
@@ -1154,8 +1154,8 @@ SimpleButton3(content, shortcut, shortcutHint, fn, style : [FormButtonStyle]) {
 		} else if (isPressed) {
 			CachedSwitch(hover_case, [content, pressed[0].form])
 		} else {
-			content
-		} |> NonZeroSize
+			content |> NonZeroSize
+		}
 	);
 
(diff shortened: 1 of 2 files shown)


lib/ui/buttons.flow
```diff
@@ -1268,8 +1268,8 @@ ImageNewButtonGeneral2(dir : string, img : string, text : string, description :
 				OnLoaded(\-> { nextDistinct(err, false) })
 			]);
 		} else {
-			pic;
-		} |> diffSkinResize |> glowImageFn;
+			pic |> diffSkinResize |> glowImageFn;
```
NikitaFil, line 1271: This one probably was intended to be applied to both branches

NikitaFil (approved):
Don't really know. I guess it is more likely that intension was to wrap both branches, especially in second case.

However, since there was no issues with it and most likely nobody really uses these elements in production - let's keep the logic as it is.

bvaschenko:
https://github.com/area9innovation/flow9/commit/3084dedc965a5ce9977cdc0779824ea18712212a

PR #1459, opened 2025-03-25
Fix memory leak in TextClip and FlowSprite

ykuzmenko73:


ykuzmenko73:
[Reader in web crashed after a few pages](https://trello.com/c/C9dcrwpM/26181-reader-in-web-crashed-after-a-few-pages)

platforms/js/TextClip.hx
```diff
@@ -205,6 +205,15 @@ class TextClip extends NativeWidgetClip {
 		style.wordWrap = false;
 		style.wordWrapWidth = 2048.0;
 
+		this.onAdded(function() {
```
NikitaFil, line 208: Looks like we can keep 'disableSprites' only for TextClip. Just subscribe on clip added and keep this subscription until it is removed.

NikitaFil (approved):
Fine like this, but can be reduced

NikitaFil (approved):
LGTM

PR #1467, opened 2025-04-03
strRepeat : tail-recursive version

shkel:


Eugeny48 (approved):
Our `int` is 32 bit, so we can get only 30 nested calls.
But tail recursion is a great thing. Thank you, @shkel.

PR #1468, opened 2025-04-04
[Material] Stepper: Make footer adaptive

mmissiato:
- https://trello.com/c/xxSNR8xG/3707-stepper-element-doesnt-adapt-to-a-panel-size-when-switches-between-steps


NikitaFil (approved):
Not sure if there is a lot of usecases where it will be helpful, but nothing harmful to include it in default fn as well. So, LGTM.

PR #1470, opened 2025-04-08
Improve `fthrottle2` by using `movableTimer`;

Eugeny48:
https://github.com/user-attachments/assets/a1b0933a-b3e1-4a43-8edd-2ce57c993765

I'm not sure if this is really useful, but in the case in the attached video it replaces 4 redraws with 2.

The hover triggers more then 5k calls to interruptibleTimer inside fthrottle2.


Eugeny48:
[OOM when trying to preview a Reader](https://trello.com/c/2pNKBY0D/27592-oom-when-trying-to-preview-a-reader)

PR #1474, opened 2025-04-23
Added the ability to set transparency for MDropDownIcon

deltaonlyone:
https://trello.com/c/oS2MqpNI/27349-change-the-font-opacity-of-the-filters-sorting-and-show-hide-on-the-dashboard


lib/material/internal/material_menu.flow
```diff
@@ -2272,6 +2272,7 @@ MDropDownMenuStyle2MMenuStyle(style : [MDropDownMenuStyle]) -> [MMenuStyle] {
 
 MDropDownIcon(opened : Transform<bool>, style : [MDropDownIconStyle]) -> Material {
 	iconColorM = tryExtractStruct(style, MWhite());
+	iconOpacity = extractStruct(style, FillOpacity(0.54));
```
NikitaFil, line 2275: I'd better use extractStructMany/tryExtractStruct in this case, so that default value declaration remains at MIcon2T.

PR #1480, opened 2025-05-03
With `dump-dead-type=1` the dead types are removed sequentially

dmitry-vlasov:
With `dump-dead-type` the dead types are removed sequentially, because the dead types eliminaiton lambda uses a common mutable list of removed types. Otherwise optimized concurrent version is used.


alstrup:
I do not completely understand what this is, but does not look it will hurt.

dmitry-vlasov:
> I do not completely understand what this is, but does not look it will hurt.

Actually it's a bug fix. By default the concurrency is on, and there's an unsafe modification of a mutable variable from different threads. That's why I separated the case (i.e. `dump-dead-types=1`), when this modification happens, and made it sequential. In normal/standard situation dead types are removed concurrently - there's no need to track the dead types.

PR #1481, opened 2025-05-03
Flowc cpp3 backend fixes

dmitry-vlasov:
The `cpp3` backend is updated and made in-sync with the current state of the standard library. Some bugs are fixed.

Now `flowc` may bootstrap itself on the `cpp3` backend, e.g. by executing following command in the `tools/flowc` directory:
```
flowc1 \
    cpp3=flowc2 \
    cpp-build-jobs=<choose appropriate number of workers> \
    cpp-verbose=1 \
    server=0 \
    cpp-compiler=clang++ \
    cpp-mimalloc=1 \
    dce-types=1 \
    \
    flowc
```
This build command assumes, that `clang++` and `mimalloc` are installed on the system. That's the most fast option, in case you use default c++ compiler and no mimalloc library, just remove appropriate lines from the command.

**NOTE:** The number of workers typically depends on the amount of RAM available: with 16Gb of RAM it's better to set it not more then 4. With 64Gb of RAM onboard you can set the number of logical cores available.


alstrup:
Approved. Very cool. How fast is the result compared to Java?

alstrup (approved):

PR #1484, opened 2025-05-07
Print an error due to a large response

Eugeny48:
Browser (Chrome Version 136.0.7103.93) does not allow to get responses longer than 512M using XMLHttpRequest().responseText

After this fix:
image

Now we can at least find the error in the console.


MalkoVl (approved):
Cool!

NikitaFil (approved):

yegordovganich (approved):

PR #1492, opened 2025-05-20
Introduce MMark

mmissiato:
highlight relevant text using mark.js


lib/material/material_mark.flow
```diff
@@ -0,0 +1,37 @@
...
+		else {
+			MGetManager(\manager -> {
+				// TODO: Use different id?
+				id = i2s(makeTagId(manager.manager));
+				querySelectorField = "markText" + "-" + id;
+				MConstruct(
+					[
+						\-> {
+							// DOM not ready on first call: Can't find element.
+							deferUntilRender(\-> hostCall("performMark", [fgetValue(text), querySelectorField])|>ignore);
```
NikitaFil, line 22: deferUntilNextFrameRendered

www/js/mark/mark-utils.js
```diff
@@ -0,0 +1,39 @@
...
+function getMarkInstance(querySelectorField) {
+	console.log('Getting Mark.js instance for selector:', querySelectorField);
+
+	const rootElement = document.querySelector('.' + querySelectorField);
+	if (!rootElement) {
+		console.warn('Root element not found for selector:', querySelectorField);
+		return null;
+	}
+
+	const textWidgets = rootElement.querySelectorAll('.textWidget');
```
NikitaFil, line 14: TODO: Add a filter for textwidgets we'd like to ignore.

www/flowjs.html
```diff
@@ -18,4 +18,6 @@
 	<script type="text/javascript" src="js/pixi.filters.js?10" defer></script>
 	<script type="text/javascript" src="js/jscommon.js?14" defer></script>
 	<script type="text/javascript" src="js/purify/purify.min.js?1" defer></script>
+	<script type="text/javascript" src="js/mark/mark.min.js" defer></script>
```
NikitaFil, line 21: It would be better to add them dynamically on first MMark usage (create a helper function in Haxe)

www/flowjspixi.css
```diff
@@ -1,6 +1,6 @@
-*:not(body):not(html) {
+/* *:not(body):not(html) {
```
NikitaFil, line 1: We should have a probation period with this part deleted under 'new=1', before merging MMark.

NikitaFil (commented):
Generally, looks good. Do not forget to make sure we have rights to use mark.js

PR #1494, opened 2025-05-30
FSelect2. Implemented optimization to prevent redundant function calls.

NikitaFil:
While working on real code, it was noticed that fselect2 implementation causes unnecessary function calls.

For example, you have beh1, beh2 and transform1 based on beh1. If you create  fselect2(transform1, beh2), subscribe on it and call fgetValue(transform1), it will call transform1 function one more time, since it wasn’t initialized on subscription. If transform1 function is heavy, it can be quite unexpected and frustrating for developers.

You can find another, richer examples in test_fselect2_optimization_1.flow

https://trello.com/c/a34GfQDg/832-fselect2-implementation-leads-to-double-calls


NikitaFil:
image
image

Eugeny48:
> > It looks good to me. But I did not understand what is wrong with tests/test_fselect2_optimization_2.flow.
> 
> If you call tests/test_fselect2_optimization_2.flow with simple fselect2(...) { FSelect2(b1, b2, fn, FDestroyed()); } it'll make double function calls.

Now I see, thank you. It prints
```

> Subscribe to last
  call fn1
  call fn2
  call fn3 with [0, 0]
  call fn4
--------------------
> Update one of the behaviours
  call fn1
  call fn3 with [1, 0]
  call fn4
  call fn2
  call fn3 with [1, 1]
  call fn4
~~~~~~~~~~~~~~~~~
> get value
1
```

alstrup (approved):

PR #1503, opened 2025-06-20
Native java function for getting the JAR time stamp

sfandersen:
The compiled timestamp of tools can be quite use full, but including the build date via "#include build_date.txt" can make builds slow. 

This will use the timestamp of the jar file. This is not the same but for tools deployed via a build server, it can be good enough.

### DIFF
--- a/lib/sys/self_timestamp.flow
+++ b/lib/sys/self_timestamp.flow
@@ -1,0 +1,25 @@
+import maybe;
+
+export {
+	// Return the ISO time stamp of the jar file. It returns none if it is not Java or not a jar file. 
+	// e.g., "2025-06-20T10:44:06Z"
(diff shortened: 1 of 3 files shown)


sfandersen:
More info about the env command line.: https://github.com/area9innovation/flow9/blob/5e9683305df103858766079ae53a4ab25badc0ff/tools/flowc/manipulation/compile_time.flow#L5-L22

sfandersen:
If the "env=" trick speeds up recompiles, then it is probably a better option.

Eugeny48:
> I see 8s on good objc cache and 32s after I only change the timestamp of that one file.

It is expected. Source file changed, so flowc recompiles all chain of files starting from the changed file up to main flow file.

PR #1508, opened 2025-07-09
Calculate the length of words in real context

AlexandrRzz:
https://trello.com/c/ZnfhB18J/28270-incorrect-words-width


lib/tropic/tropic_paragraph.flow
```diff
@@ -962,6 +986,42 @@ commonWordConstructor(
 	);
 }
 
+calculateWordTextAndMetrics(
```
NikitaFil, line 989: It's not so obvious that calculateWordTextAndMetrics both updates inspector and returns collected data. I'd add a comment to prevent confusion.

NikitaFil (approved):
In general, looks good.

PR #1513, opened 2025-07-25
uploadNativeFile: return error description instead of just code

pzotov:
Task https://trello.com/c/mJNZrFk8/27510-there-is-no-information-why-the-file-could-not-be-upload


MalkoVl (approved):

NikitaFil (approved):
I'm fine with it

PR #1529, opened 2025-09-16
Native `db2time`

Eugeny48:
We have a lot of time fields in db, so the speed of parsing these fields matters.

Card: https://trello.com/c/7KsGgL8m/1696-slow-makessqlupdaterequest

Native function results exactly like flow `db2time` and twice faster in JS:
image

Flow function:
image

Native function:
image

### DIFF
--- a/lib/date.flow
+++ b/lib/date.flow
@@ -1,8 +1,4 @@
 // © Copyright 2011 Area9 Technologies.
-
-import math/math;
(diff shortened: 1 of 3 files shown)


Eugeny48:
[Slow deserializeSTime](https://trello.com/c/7KsGgL8m/1696-slow-deserializestime)

PR #1531, opened 2025-09-19
fixed announcing the number of menu items using SR

SLemishko:
https://trello.com/c/PhIJb0Av/28500-coach-submenu-issues


NikitaFil (approved):
Good. But please add to commit description that we are trying to enable screenreader list items order announcing, which is probably was disable by previous commits. This is due to PO request in the card.

PR #1534, opened 2025-09-26
JS: add app badge count and handling for notification badge field

DMLatrofa:


vereschagai (approved):
Generally OK, but the PR doesn't provide an ability to increment badge counter while application is in background. For that it needs to be implemented with Firebase service worker.

DMLatrofa:
> Generally OK, but the PR doesn't provide an ability to increment badge counter while application is in background. For that it needs to be implemented with Firebase service worker.

The best method is to have the badge count value directly in the request. When we have "inc" to increment the count we need to have a background listner in the application so that the js client that has access to the local storage is called and can properly increment the count. The alternative to have the service worker handle this is to implement an IndexedDB or use Cache Storage

PR #1537, opened 2025-09-29
add id generator

SLemishko:


lib/ds/id_generator.flow
```diff
@@ -0,0 +1,37 @@
+import maybe;
+import ds/tree;
+import string;
+
+export {
+	createStringIdGenerator(prefix : string) -> IdGenerator;
```
NikitaFil, line 6: Add comment about  how to use

lib/ds/id_generator.flow
```diff
@@ -0,0 +1,37 @@
...
+	generateIntId() -> int;
+}
+
+intIdR = ref 0;
+
+generateIntId() -> int {
```
NikitaFil, line 12: Do I understand correctly, that intIdR is shared between all id generators and generateIntId, called from different usages, will generate unique ids? Sometimes it can be undesired behavior, since some ids at one specific usage would be missed.

lib/material/internal/material_dialog_internal.flow
```diff
@@ -842,13 +843,7 @@ renderMSnackbar(manager : MaterialManager, text : string, style : [MSnackbarStyl
...
 }
 
-snackBarIdCounterR = ref 0;
-
-getSnackBarIdAndUpdateCounter() -> string {
-	currentId = ^snackBarIdCounterR;
-	snackBarIdCounterR := currentId + 1;
-	"snackeBarId_" + i2s(currentId)
-}
+getNextSnackBarIdFn = createStringIdGenerator("snakBarId").nextId;
```
NikitaFil, line 846: typo : 'snackbar'

PR #1539, opened 2025-09-30
matchesWithGlobPattern : infinite recursion

shkel:
The Problem :
The infinite recursion occurred because when the string was empty and the glob started with *, the function would call itself with the same empty string:
`if (strlen(str) == 0) {
	if (f == "*") matchesWithGlobPattern(str, strRight(glob, 1))  // str still ""`

The Solution :

1. Check glob first: If the glob is empty, we only match if the string is also empty
2. Extract restGlob once: Calculate strRight(glob, 1) once and reuse it to avoid redundant calls
3. Proper base case: When both string and glob are empty, return true; when glob is empty but string isn’t, return false

The key fix ensures that when processing * with an empty string, we advance the glob pattern (removing the *) before recursing, which eventually reaches the base case where the glob becomes empty.
The function now correctly handles cases like:
matchesWithGlobPattern("", "*") → true
matchesWithGlobPattern("", "**") → true
matchesWithGlobPattern("", "*a") → false
matchesWithGlobPattern("test", "*") → true

### DIFF
--- a/lib/text/wildcards.flow
+++ b/lib/text/wildcards.flow
@@ -67,17 +67,30 @@ isGlobPattern(str : string) -> bool {
 }
 
 matchesWithGlobPattern(str : string, glob : string) -> bool {
-	f = strLeft(glob, 1);
-	if (strlen(str) == 0) {
-		if (f == "*") matchesWithGlobPattern(str, strRight(glob, 1))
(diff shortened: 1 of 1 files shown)


shkel:
merged https://github.com/area9innovation/flow9/commit/47733643d884068deaf836d92a9da18795c7c637

PR #1604, opened 2026-04-30
MExpander: introduce MExpanderAriaLabelledByHeader

ykuzmenko73:
https://trello.com/c/Ak9b1Wnt/29305-add-shortcut-instruction-for-editor

currently we have the following HTML for expander's button:

```
<button id="_1x8ht6uvn" class="nativeWidget accessButton" aria-controls="expander_content_1" aria-describedby="expander_header_1" aria-expanded="false" aria-label="Expand" role="button" style="visibility: visible; width: 42px; height: 42px; pointer-events: auto;" tabindex="124"><p id="_6crzxlh38" class="nativeWidget textWidget" role="presentation" style="visibility: visible; white-space: pre; color: rgba(117, 117, 117, 0.8); font-family: &quot;Material Icons&quot;; font-size: 24px; line-height: 28px; width: 24px; height: 28px; left: 9px; top: 7px;"></p></button>
```

we have to remove aria-label="Expand"  and replace aria-describedby="expander_header_1" to aria-labelledby="expander_header_1"

it would be good to remove role="button" too, but this PR does not do it.


lib/material/internal/material_component.flow
```diff
@@ -549,10 +549,10 @@ MComponent2TShortcut(manager : MaterialManager, content : Tropic, state : [flow]
 
 MComponent2TAction(manager : MaterialManager, content : Tropic, state : [flow], focus : MFocus, keyOnClick : Maybe<(bool) -> void>, shortcut : string) -> Tropic {
 	keepFocusOnDown = extractStruct(state, MKeepFocusOnDown(ref false)).enabled;
+	buttonTitleTransform = extractStruct(state, MButtonTitle(const(toCapitalization(extractStruct(state, AccessRole("button")).role)))).title;
 	properties =
-		arrayPush(
-			replaceStruct(extractFAccessProperties(state), extractStruct(state, AccessRole("button"))),
-			FAccessAttribute("aria-label", extractStruct(state, MButtonTitle(const(toCapitalization(extractStruct(state, AccessRole("button")).role)))).title)
+		(\p -> if (fgetValue(buttonTitleTransform) != "") arrayPush(p, FAccessAttribute("aria-label", buttonTitleTransform)) else p)(
```
NikitaFil: Do not clearly understand why should we filter out empty button title? If it appears here, then it has to be intentional.

NikitaFil (commented):
Conceptually, fine with it. I'd even like to make this behaviour by default, without creating additional style. The only limitation is empty / specific header cases, which should be supplied with fallback labels.

BTW, you can set tagname and role separately and directly, but I dont understand why it might be beneficial to remove button role.

NikitaFil (approved):
I'm fine with it

PR #1605, opened 2026-05-04
Sha256

bvaschenko:
https://trello.com/c/911LwgBg/30126-internal-security-audit-f7-weak-cryptographic-hash-md5-for-request-signing


yegordovganich:
@review

yegordovganich:
# Claude review:

## Overall

Good work. This directly addresses **F7** by providing the Flow-side crypto primitives (`sha256`, `hmacSha256`, AES-256-GCM) needed to replace MD5 request signing. The Java implementation is solid, the test coverage is thorough, and the PHP cross-platform test is a smart inclusion. A few issues to address:

---

## Issues

### 1. **`crypto_aes.flow` — Fallback function parameter names are wrong**
```flow
encryptAesGcm(binarySource: string, algorithm: string) -> string {
decryptAesGcm(binarySource: string, algorithm: string) -> string {
```
These should be `(data: string, key: string)` to match the native declarations above. The names `binarySource`/`algorithm` are copy-paste artifacts and will confuse anyone reading the fallback.

### 2. **`crypto_aes.flow` — Empty string is ambiguous as error sentinel**
Both `encryptAesGcm` and `decryptAesGcm` return `""` on error. But empty string is a valid plaintext — the test suite even tests empty plaintext round-trips. This means callers **cannot distinguish** "decryption failed" from "successfully decrypted empty plaintext." Consider using a `Maybe<string>` return type, or document that empty plaintext must never be used.

### 3. **JS target missing AES-256-GCM implementation**
`FlowCrypto.hx` only implements `sha256` and `hmacSha256`. If `crypto_aes.flow` is imported from a JS target, it silently falls back to returning `""` (the Flow fallback). This is a silent failure trap. Either:
- Add AES-256-GCM to the Haxe/JS target (using `js.lib.Crypto`), or
- Add a comment/runtime warning that AES is Java-only

### 4. **`FlowCrypto.java` — Error output goes to stdout**
```java
System.out.println("sha256 exception: " + e.getMessage());
System.out.println("hmacSha256 exception: " + e.getMessage());
```
These should go to `System.err` or use a logger. Printing to stdout pollutes program output and can break the cross-test (which parses stdout).

### 5. **Test file paths in comments are wrong**
`test_crypto_aes.flow` line 12 says:
```
flowc1 lib/test_crypto_aes.flow jar=test_crypto_aes.jar
```
But the file is at `tests/test_crypto_aes.flow`. Same for the PHP cross-test reference on line 22. The PHP file header has the same issue (`lib/test_crypto_aes_php_cross.php` vs `tests/`).

### 6. **`sha256.flow` — No Flow fallback**
Unlike `crypto_aes.flow` which provides fallback stubs, `sha256.flow` declares bare natives with no fallback. If the native host isn't available, this fails at link time. Should either add fallback stubs (like `crypto_aes.flow` does) or document this is intentional.

### 7. **`FlowCrypto.hx` — Missing trailing newline**
Minor, but the file lacks a trailing newline, which some tools flag.

---

## Nits (non-blocking)

- **Key derivation via bare SHA-256** (`deriveKey()`) is fine for high-entropy hex keys but is NOT a proper KDF (no salt, no iterations). Worth a comment that this assumes the input key already has sufficient entropy.
- The `encryptAesGcm`/`decryptAesGcm` Java implementations are clean and correct — proper IV generation with `SecureRandom`, explicit tag separation, correct GCM parameter spec.
- The PHP cross-test is well-structured with `escapeshellarg` on all inputs. 👍

---

## Verdict

**Request changes** — issues #1 (wrong param names), #4 (stdout pollution), and #5 (wrong paths) are straightforward fixes. Issue #2 (empty string ambiguity) and #3 (missing JS AES) are design decisions worth resolving before merge.

@bvaschenko FYI

PR #1612, opened 2026-05-14
Optimize switch code generation: preserve union-level cases from desugar to backends

Eugeny48:
### Problem

When a Flow `switch` uses union names in cases, the desugarer previously expanded each union case into individual struct cases, duplicating the case body for every struct in the union. For nested switches over large unions, this caused **O(n×m) code explosion** — the same body was repeated dozens of times in generated Java and JavaScript output.

## Example
```flow
import server/ssql/ssql_i;

main() {
	t : STableUnion = STable(["field"], []);
	q : SSql = t;

	switch (q) {
		SValue(): {
			switch (q) {
				SSimpleValue(): {
					println("The query is a simple value: " + q.structname);
				}
				STableUnion(): {
					println("The query is a table with fields: [" + strGlue(q.fieldNames, ", ") + "]");
				}
				default: {
					println("The query is a value, but not simple: " + toString(q));
				}
			}
		}
		default: {
			println("The query is not a value: " + q.structname);
		}
	}

	println(t.fieldNames);
}
```

flowc translates this flow code to such JS code
```js
function main(){
  var t=c$F3(["field"],[]),q=t;
    (function(){var s$;switch(q._id){
    case 272:{s$=(function(){var s$;switch(q._id){
        case 272:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 198:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 248:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 181:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 228:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 284:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 152:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 214:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 211:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 289:{s$=console.log(HaxeRuntime.toString((("The query is a table with fields: ["+A9__strGlue(q.fieldNames,", "))+"]"), true));break;}
        case 290:{s$=console.log(HaxeRuntime.toString((("The query is a table with fields: ["+A9__strGlue(q.fieldNames,", "))+"]"), true));break;}
        default:{s$=console.log(HaxeRuntime.toString(("The query is a value, but not simple: "+A9__toString(q)), true));break;}
        }return s$;}());break;}
    case 198:{s$=(function(){var s$;switch(q._id){
        case 272:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 198:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 248:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 181:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 228:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 284:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 152:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 214:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 211:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 289:{s$=console.log(HaxeRuntime.toString((("The query is a table with fields: ["+A9__strGlue(q.fieldNames,", "))+"]"), true));break;}
        case 290:{s$=console.log(HaxeRuntime.toString((("The query is a table with fields: ["+A9__strGlue(q.fieldNames,", "))+"]"), true));break;}
        default:{s$=console.log(HaxeRuntime.toString(("The query is a value, but not simple: "+A9__toString(q)), true));break;}
        }return s$;}());break;}
...
```

### Solution

Introduce `FiCaseUnion` — a new IR node that preserves union-level switch cases throughout the compilation pipeline, instead of expanding them eagerly during desugaring.

```flow
FiSwitchCase ::= FiCase, FiCaseUnion;
FiCase(struct: string, argNames: [string], body: FiExp, start: int);
FiCaseUnion(union: string, structs: [string], body: FiExp, start: int);
```

### Commits

1. **Java backend: merge switch cases with shared fields using Field_ accessor interfaces** — When multiple struct cases in a switch access the same fields, merge them into a single case using `Field_` accessor interfaces instead of duplicating the body.

2. **Introduce FiCaseUnion** — Core infrastructure: new `FiSwitchCase` union type, desugar produces single `FiCaseUnion` per union case (body desugared once), type checker handles union cases, all 50+ backend and manipulation files adapted using `fiExpandSwitchCases` helper for backward compatibility.

3. **Java backend: native FiCaseUnion support** — Emits Java fall-through `case` labels for each struct in the union, with a single shared body. Uses `Field_` accessor interfaces (via `generalizeVarType` → `FiTypeFlow()`) for field access across struct types.

4. **JavaScript backend: native FiCaseUnion support** — Emits JS fall-through `case` labels similarly. Adds `fiJsSwitchCaseLabels` helper. Adapts `isIf` optimization and `>=50` case heuristic for `FiCaseUnion`.

5. **Update flowc.jar with FiCaseUnion support**

6. **Fix overlapping union/struct cases in switches** — Switches can have both a union case (e.g. `CharacterStyle()`) and a struct case for a member of that union (e.g. `SetRTL()`). Added dedup in `solve_expectations` and struct subtraction in `fc2fi` to handle overlaps.

7. **Fix type checker bugs for FiCaseUnion switch cases** — Fix union case type parameter instantiation in desugar, fix `names2unions` expansion for union names, fix `bestunion` lookup for union cases.

8. **Fix type checker: common parent union checks and union-struct narrowing** — When unnamed unions contain structs from a common parent union, suppress false type errors. Fix `FUnion ≤ FStruct` and `FStruct ≤ FUnnamedUnion` containment paths.

9. **Fix type checker: detect incompatible type annotations as errors** — Type annotations like `x : [SubType] = expr` where `expr` has type `[SuperType]` now produce errors instead of silently accepting the incompatible narrowing.

10. **Fix type checker: tyvar contamination through polymorphic calls and union case switch type** — Fix union case switch to use `unionType` instead of `varType`. Remove gl2 guard in `funifyBounded` to prevent type variable contamination through `grow_right` in non-bounded FUnifyLeft path.

11. **Fix type checker: restore gl1 guard for bounded FUnifyLeft path** — The gl1 guard in `funifyBounded` is needed for the bounded+FUnifyLeft path (fixes Maybe/Some narrowing in switch cases). Only gl2 removal was correct.

12. **Update flowc.jar**

### Key design decisions

- **Clean semantic separation**: `FiCase` = single struct with field bindings; `FiCaseUnion` = union with list of structs, no field bindings, body uses union-typed variable
- **Backward compatible**: backends that don't need native support call `fiExpandSwitchCases()` to get `[FiCase]` as before
- **No expand-then-merge**: removed the old `fiMergeExpandedUnionCases` pass that did wasteful double work (expand all cases, then try to re-merge identical bodies)

### The same example after the PR

```js
function main(){
  var t=c$F3(["field"],[]),q=t;
    (function(){var s$;switch(q._id){
    case 272:
    case 198:
    case 248:
    case 181:
    case 228:
    case 284:
    case 152:
    case 214:
    case 211:
    case 282:
    case 289:
    case 290:{s$=(function(){var s$;switch(q._id){
        case 272:
        case 198:
        case 248:
        case 181:
        case 228:
        case 284:
        case 152:
        case 214:
        case 211:{s$=console.log(HaxeRuntime.toString(("The query is a simple value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
        case 289:
        case 290:{s$=console.log(HaxeRuntime.toString((("The query is a table with fields: ["+A9__strGlue(q.fieldNames,", "))+"]"), true));break;}
        default:{s$=console.log(HaxeRuntime.toString(("The query is a value, but not simple: "+A9__toString(q)), true));break;}
        }return s$;}());break;}
    default:{s$=console.log(HaxeRuntime.toString(("The query is not a value: "+(HaxeRuntime.getStructName(q._id))), true));break;}
    }return s$;}());
    return console.log(HaxeRuntime.toString(t.fieldNames, true));
}
```

### Code size reduction

`educator.js` compiled with flag `debug=1`: **61.0 MB → 59.4 MB** (−1.6 MB, **−2.7%**).

### Adding native FiCaseUnion support to other backends

Currently only Java and JS backends emit compact fall-through code for `FiCaseUnion`. All other backends (C++, bytecode, D, Lisp, ML, Nim, Wasm, Wise) use `fiExpandSwitchCases()` to expand union cases back to individual struct cases before code generation — they produce correct but unoptimized output.

To add native support to another backend, follow the Java/JS pattern:

1. In the `FiSwitch` code generation, iterate `[FiSwitchCase]` directly instead of calling `fiExpandSwitchCases`
2. Handle `FiCaseUnion` by emitting fall-through case labels for each struct in `FiCaseUnion.structs`, with the body emitted once
3. Handle `FiCase` as before (single struct with field bindings)
4. For field access in `FiCaseUnion` bodies: the switch variable has union type, so use the target language's equivalent of interface-based dispatch (Java uses `Field_` accessor interfaces; other backends may use casts to a common base type or dynamic dispatch)

### DIFF
--- a/tools/flowc/backends/bytecode/fi2bytecode_compile.flow
+++ b/tools/flowc/backends/bytecode/fi2bytecode_compile.flow
@@ -257,7 +257,8 @@ fiBcEncode(cfg : FcBytecodeConfig, ctx : FiBcScopeContext, expr : FiExp, tailcal
 				elsecode
 			], pc);
 		}
-		FiSwitch(e0, switchType, cs, type, i): {
+		FiSwitch(e0, switchType, cs0, type, i): {
+			cs = fiExpandSwitchCases(cs0);
 
 			arg = fiBcEncode(cfg, ctx, e0, false, pc);
 
--- a/tools/flowc/backends/cpp/fc2cpp_compile.flow
+++ b/tools/flowc/backends/cpp/fc2cpp_compile.flow
@@ -8,6 +8,7 @@ import tools/flowc/manipulation/find_last_varrefs;
 import tools/flowc/manipulation/common;
 import tools/flowc/manipulation/freevars;
 import tools/flowc/incremental/fiprogram;
+import tools/flowc/incremental/fi_helpers;
 import ds/set;
 
 export {
@@ -611,7 +612,8 @@ fiCppCompileToExpr(
 			}
 		}
 
-		FiSwitch(e0, e0type, cs, __, __): {
+		FiSwitch(e0, e0type, cs0, __, __): {
+			cs = fiExpandSwitchCases(cs0);
 			aVar = fiCppCompileToExpr(ctx, e0, indent, false, false);
 
 			isUnion = fiIsUnion(program, fiExpType(e0));
--- a/tools/flowc/backends/cpp2/fi2cpp_compile.flow
+++ b/tools/flowc/backends/cpp2/fi2cpp_compile.flow
@@ -865,7 +865,8 @@ cpp2WriterCompileExpression(writer : Cpp2Writer, code : FiExp, ctx : Cpp2Context
 			cpp2ContextWr(ctx, "}");
 			cpp2ContextJoin(ctx, [ctx_if, ctx_else]);
 		}
-		FiSwitch(e0, sw_type, cases, type, __): {
+		FiSwitch(e0, sw_type, cases0, type, __): {
+			cases = fiExpandSwitchCases(cases0);
 			mvexpr = cpp2WriterGetValueRef(writer, e0, ctx, false);
 			vexpr = switch(mvexpr) {
 				Some(ve): ve;
--- a/tools/flowc/backends/cpp3/fi2cpp3_util.flow
+++ b/tools/flowc/backends/cpp3/fi2cpp3_util.flow
@@ -385,8 +385,9 @@ fiExp2stringFull(expr: FiExp) -> string {
 			(if (sl_test(e2_s)) e2_s else "{\n" + strIndent(e2_s) + "\n}") + " else " +
 			(if (sl_test(e3_s)) e3_s else "{\n" + strIndent(e3_s) + "\n}");
 		}
-		FiSwitch(x, __, cases,__, __): {
-			"switch (" + x.name + ") {\n" + 
+		FiSwitch(x, __, cases0,__, __): {
+			cases = fiExpandSwitchCases(cases0);
+			"switch (" + x.name + ") {\n" +
 				strIndent(superglue(cases, \c -> {
 					body_s = fiExp2stringFull(c.body);
 					body_s1 = if (sl_test(body_s)) {
--- a/tools/flowc/backends/d/fi2d_compile.flow
+++ b/tools/flowc/backends/d/fi2d_compile.flow
@@ -2,6 +2,7 @@ import string_utils;
 import tools/flowc/backends/d/fi2d_utils;
 import tools/flowc/backends/d/fi2d_utils2;
 import tools/flowc/backends/d/fi2d_assemble;
+import tools/flowc/incremental/fi_helpers;
 
 export {
 	fiDCompileToplevel(
@@ -426,7 +427,8 @@ fiDCompileToExpr(cfg : FiDConfig, program : FiProgram, expr : FiExp, indent : st
 			p1 = (if (prefix != "") "new " + prefix else "");
 			"/*CAST!"+toString(tFrom)+"*/" + p1 + "(cast(" + tt + ")" + ex + ")";
 		}
-		FiSwitch(e0, e0type, cs, __, __): {
+		FiSwitch(e0, e0type, cs0, __, __): {
+			cs = fiExpandSwitchCases(cs0);
 			aVar = fiDCompileToExpr(cfg, program, e0, indent, isVoid);
 			defCase = filter(cs, \e -> e.struct == "default");
 			dc = if (defCase == []) " else {assert(false, \"" +aVar + "\");}" else {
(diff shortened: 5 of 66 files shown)


alstrup:
@Eugeny48 If it works, then LGTM.

Eugeny48:
> @Eugeny48 If it works, then LGTM.

It does not work yet, it require one more fix in type checker. I'm working on it.

Eugeny48:
@alstrup Now PR is ready. The type checker is updated. It’s still not perfect, just a bit better.

PR #1626, opened 2026-06-30
Migrate JWT handling from auth0 java-jwt to JJWT (drop Jackson)

Eugeny48:
## Why
Jackson is flagged by CVE-2026-54515 (and CVE-2026-54512). It is not used directly
anywhere here — only a transitive dependency of com.auth0:java-jwt, used by FlowJwt.
Jackson 3.x isn't supported by java-jwt and uses incompatible tools.jackson coordinates,
so this migrates JWT to io.jsonwebtoken:jjwt with the Gson backend (already on the
classpath), removing Jackson from the build entirely.

## What
FlowJwt.java: com.auth0.jwt.* → io.jsonwebtoken.* (JJWT 0.13). Sign/verify/decode entry
points, HS/RS/ES 256/384/512, kid header and 300s leeway all preserved; decodeJwt still
returns the raw payload. Claims parsed via existing json-simple to keep number types.
Key-parsing helpers untouched.

## ⚠️ Behavioural change — HMAC key length
auth0 accepted HMAC secrets of any length; JJWT enforces the RFC 7518 minimum
(HS256≥32B, HS384≥48B, HS512≥64B) and now throws a descriptive WeakKeyException.
Verify HS* secrets meet the minimum.

### DIFF
--- a/platforms/java/com/area9innovation/flow/FlowJwt.java
+++ b/platforms/java/com/area9innovation/flow/FlowJwt.java
@@ -18,13 +18,16 @@ import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.json.simple.*;
 import org.json.simple.parser.*;
 
-import com.auth0.jwt.JWTVerifier.*;
-import com.auth0.jwt.interfaces.*;
-import com.auth0.jwt.exceptions.*;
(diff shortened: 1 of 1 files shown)


bvaschenko (approved):
LGTM

PR #1638, opened 2026-07-27
MMenu. Fixed MSubMenuDynamic with MExpanding in menus with more than 100 rows

VladislavFedurko:
## Problem

`MSubMenuDynamic` with `MExpanding()` breaks when the menu popup contains more than 100 rows: closed submenus appear with blank gaps sized like a previous submenu's children, and clicking a broken submenu renders its items at the wrong offset.

## Root cause

With more than 100 rows, `MPopup` uses `MRecyclerGrid2T` for performance. The recycler caches each row's measured size and lays out row positions from those cached values. Expanding submenus change row heights at runtime (`MShow` toggle), but rows that are not currently rendered are never re-measured, so subsequent rows are positioned using stale sizes.

## Fix

Menus that contain a submenu with `MExpanding()` style now use regular `MScroll` instead of `MRecyclerGrid`:

- `material.flow`: added `RMNoRecyclerGrid()` to `MPopupStyle`.
- `material_popup.flow`: recycler path is skipped when `RMNoRecyclerGrid()` is set.
- `material_menu.flow`: `MMenuLines2MPopup` detects `MExpanding()` submenus and sets the flag.

Only menus with expanding submenus are affected; all other popups keep using the fast recycler path.

### DIFF
--- a/lib/material/internal/material_menu.flow
+++ b/lib/material/internal/material_menu.flow
@@ -2212,6 +2212,15 @@ MMenuLines2MPopup(
 	labelIdx = b2i(isSome(mLabel)) + b2i(useFocusInspector);
 	itemsLen = length(newItems);
 
+	// MRecyclerGrid caches row sizes and does not handle rows that change
+	// height at runtime (e.g. MExpanding submenus), so use regular scroll then
+	hasExpandingSubMenu = exists(items0, \it -> switch (it : MMenuLine) {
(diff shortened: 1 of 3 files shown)


VladislavFedurko:
I couldn't fix it myself, so I used Mindcode
Before:
image
After:
image

stanislavpodolia (approved):

PR #1644, opened 2026-08-04
fix(flow9): Add window management natives for popup-safe navigation

vadlis:
Add three new Flow natives for M16 federated PDF launch:
- openBlankWindow() : Int — open tab synchronously in gesture context
- setWindowUrlById(handle : Int, url : String) : Bool — navigate from async callback
- closeWindowById(handle : Int) : Void — cleanup on error

Solves browser popup blocker blocking window.open() from async callbacks. Enables Elsevier/Bookboon PDFs to load in Rhapsode apps.

Implementation: platforms/js/Native.hx (Haxe, JS target)
Declarations: lib/net/url.flow (with non-JS fallback bodies)


lib/net/url.flow
```diff
@@ -12,11 +12,35 @@ export {
...
 	// Opens a new window with url or redirects current window to url (if target == "_self")
 	// See possible values of target in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target
 	getUrl : (url : string, target : string) -> void;
 	getUrlAutoclose : (url : string, target : string, delay : int /*ms*/) -> void;
 
+	// Popup-safe open helpers (see native declarations above).
+	openBlankWindow : () -> int;
+	setWindowUrlById : (handle : int, url : string) -> bool;
+	closeWindowById : (handle : int) -> void;
+
```
bvaschenko: these declarations look excessive

platforms/js/Native.hx
```diff
@@ -3009,6 +3009,76 @@ Native.memoryLeakReset();
...
+		#if (js && !flow_nodejs)
+		try {
+			// No "noopener": we must retain the reference to navigate it later.
+			var w = Browser.window.open("about:blank", "_blank");
+			if (w == null) return 0;
+			var id = openedWindowsNextId++;
+			openedWindows.set(id, w);
+			return id;
+		} catch (e:Dynamic) {
+			if (e != null && e.number != -2147467259) throw e;
```
bvaschenko, line 3035: add similar comment please here and later
// Catch exception that tells that window wasn't opened after user chose to stay on page

bvaschenko (approved):
Looks good to me, 2 minor comments.

PR #1658, opened 2026-09-17
M24: global postMessage handler forwards message origin, drops source…

deltaonlyone:
# M24: Global postMessage handler forwards iframe message data into flow

https://eu.area9studio.com/nexus/lyceum-product-development/3iQvcFLoGDm-security-audit-m24-global-postmessage-handler-forwards-iframe-message-data-into-flow-without-origin

## Summary

The global `postMessage` bridge in `RenderSupport.hx` forwarded iframe message payloads into
flow while discarding `event.origin`, so no flow-side consumer had any way to tell who sent a
message (CWE-346). This PR forwards the origin alongside the data, rejects messages with no
source window, and updates the flow-side consumers to accept the new message shape.

### `platforms/js/RenderSupport.hx`

**1. Drop sourceless messages** (defense-in-depth, ~line 1394)

A genuine message forwarded to one of our iframes must carry a source window (`e.source`).
Messages without one can never match a nested iframe window, so they are rejected early —
before entering the forward path:

```haxe
if (content_win == null) return;
```

**2. Forward `e.origin` together with the message data** (~line 1419)

```haxe
// before
callflow(["postMessage", e.data]);
// after
callflow(["postMessage", e.data, e.origin]);
```

Flow consumers can now validate the message source. Per-channel origin checks deliberately
live in the consumers / wrapper HTMLs, not here.

**Foreign `e.origin` is intentionally NOT rejected at this level.** Embed-Website (plain
`MRealHTML`) and `MRealHTMLCrossDomain3` legitimately embed cross-origin content; a blanket
check here would break both.

### `lib/material/material_ui.flow` — `MRealHTMLCrossDomain3` (~line 548)

```flow
// before
} else if (length(args) == 2) {
// after
} else if (length(args) >= 2) {
```

`flowCallBack` now accepts both the legacy 2-arg form `["postMessage", data]` and the new
3-arg form `["postMessage", data, origin]`.

## Why this second change is not optional

The original M24 report claimed the extra argument was *"backward-compatible: existing
callflow handlers ignore extra args"*. **That was wrong.** `flowCallBack` used a strict
`length(args) == 2`, so adding a third argument made every forwarded message silently
vanish — no error, no log.

Concrete breakage: the SCORM player's `getInitialState` handshake never completed, and
SCORM questions rendered as a **white screen**. The `>= 2` relaxation is what makes the
RenderSupport change shippable at all.

**Note for future work:** any change that forwards additional arguments through `callflow`
must check these handlers first.

## Follow-up 1 after review: three more handlers had the same strict arity check

`material_ui.flow` was not the only `flowCallBack` requiring exactly two arguments. Review
found three more, all of which would silently drop every forwarded message:

- `lib/form/realhtml.flow` — `RealHTMLCrossDomain2` (live caller: `mediaconverter.flow`)
- `lib/tropic/trealhtml.flow` — `TRealHTMLCrossDomain2`
- `lib/material/internal/material_player.flow` — the YouTube player's `receiveB`
  back-channel feeding `fselectFromStateM`, i.e. the player state readback used by
  `MVideoPlayer` across Rhapsode (video-clip and federated questions)

All relaxed to `>= 2`; no `== 2` postMessage arity check remains in `lib/`. The comment in
`RenderSupport.hx` that called the extra argument "backward-compatible" was inherited from
the original report and is wrong — it now states the opposite explicitly and lists the
handlers, so the next person forwarding an argument checks them first.

## Follow-up 2 after review: the iScript channel is now gated

`lib/material/iscript/material_iscript.flow:636` consumed forwarded messages without any
sender check. This is the **highest-impact consumer in the codebase**: the channel
synthesises recorded user input back into the running application, so an attacker able to
post into it can drive the UI as the user.

```flow
if (origin != "" && startsWith(getBrowserURL(), origin + "/")) {
```

Plus `import sys/history` for `getBrowserURL()`.

Two notes on the shape of this check:

- `getBrowserURL()` returns the **full URL including path**, so this is `startsWith`, not
  equality — an equality test would reject every legitimate message.
- An **empty origin is rejected**. The first version of this gate accepted it (`origin == ""`)
  on the grounds that non-JS targets supply no origin — but review correctly pointed out that
  the whole `MConstruct` sits under `if (js)` (line 634), so that branch is unreachable on the
  only target where the listener exists, and it left an unconditional acceptance path for no
  reason. Fixed in `74f19b677` by flipping to `origin != "" && ...`. Written as an explicit
  test rather than simply deleting the clause: with the clause gone, an empty origin would
  reduce to `startsWith(url, "/")`, which is false only *incidentally* (URLs start with
  `http`). The explicit form states the intent.

## Follow-up 3 after review: the YouTube player frame is now gated

`www/yt_player.html:130` consumed forwarded messages without any sender check. The first
version of this PR argued the frame had to stay open because the YouTube iframe API posts
from a **child** frame. Review challenged that, and re-reading the handler showed the
argument was an unverified assumption:

```javascript
window.addEventListener("message", function (e) {
	if (e.source !== window.parent) return;
	try {
		var data = JSON.parse(e.data);
		if ('loadVideo' in data) { ... }
		if ('playerState' in data) { ... }
```

The handler consumes exactly two keys, `loadVideo` and `playerState`. Both are sent by flow
from `window.parent` (`lib/material/internal/material_player.flow:998` and the `sendState`
calls at `:1036`, `:1045`, `:1051`, delivered through `WebClip.hostCall`, which invokes
`postMessage` on the frame from the parent document). Messages from the child YouTube frame
carry neither key and fall through both `if`s with no effect — the YouTube iframe API has
its own listener, registered by the `iframe_api` script, which this guard does not touch.

The guard is placed **before** `JSON.parse` rather than after: foreign messages are never
parsed at all, which also removes the `SyntaxError` console noise from third parties that
post non-string messages.

Impact of leaving it open: any window with a handle to the frame could drive `seekTo` /
`playVideo` on the learner's video, and Rhapsode records evidence against video time ranges.

## Scope: the one consumer left ungated

The original report enumerated four `addMessageEventListener` consumers. A wider audit found
**seven shipped consumers**. One is intentionally left open, and the reason is worth
recording so the next reviewer does not "fix" it:

- **`lyceum/wigi/recursives/openpage_link.flow`** — its iframe navigates to an external LTI
  reader (VitalSource / BOOST) through `integration/openpage_service.php`. The
  `boost-primary-close` message legitimately arrives from a third-party origin we cannot
  know in advance; a same-origin gate would break the legitimate close. Flow's
  `addMessageEventListener` does not expose `event.source` at all, so no sender check is
  available either, and the only effect is closing a dialog.

Already guarded before this work, and untouched: `presenter_view.flow:54`,
`pdfviewer.html:31` (ENT-H4), `certificate_editor_wrapper.js:407`, `graphviz.html:118`,
`competency_framework_importer.html:113`, `integration/scorm/sco.js:166`,
`collins_activity.html:17`, `mail_editor_wrapper.js:182`,
`cms/page_editor/www/js/custom.js:96`, and the `prime_html` / `prime` / `cms`
`html_preview` wrappers.

Non-shipped code (editors, experiments, tests, vendored workers) was deliberately not
touched.

### Verified at runtime

- **SCORM via simulation** — course loads; full `getInitialState` → state reply →
  `LMSInitialize` → `LMSSetValue` → `LMSCommit` sequence confirmed in console.
- **Progress persistence** — re-opening a SCORM question restores saved progress.
- **Lesson status** — passing SCORM content updates lesson status correctly.
- **Embed-Website** (cross-origin iframe, plain `MRealHTML`) — still renders and responds.
- **Learner playback** — regular course playback, no regressions.
- **Build artifact inspected** — the compiled `receiveWindowMessage` in the shipped
  `curator.html.js` contains both the `content_win == null` guard and the 3-argument
  `callflow`, confirming the Haxe change actually reached the bundle.
- `flowc1` compiles clean.

### Not verified at runtime — reviewer should be aware

Honest status rather than a blanket "tested":

- The **mediaconverter path** through `realhtml.flow` / `trealhtml.flow` was verified by
  inspection and compilation only. The failure mode of a missed arity check is a **silent
  drop with no console error**.

### Verified at runtime after rebuild

- **YouTube playback** — video loads and plays with the new `e.source` guard in place; the
  `loadVideo` / `playerState` traffic from flow is unaffected.
- **iScript gate** — exercised in the running application after rebuild.
- **Embed Website block** — renders and scrolls; the `contentHeight` back-channel through
  the relaxed `flowCallBack` handlers works.
- **Presenter view** — presentation start, next / previous and stop all work through the
  gated channel (companion `lyceum` PR).
- The compiled bundle was confirmed to contain both halves of `5c24a3e9b`
  (`if (content_win == null) return;` and the three-argument `callflow`).

### Pre-existing console noise (not introduced here, not fixed here)

- `SecurityError: ... hasNestedWindow` — `receiveWindowMessage` walks every iframe looking
  for the sender and touches `contentWindow.document`, which throws for cross-origin frames.
  Caught and ignored, but printed. Harmless, just loud.
- `Error in hostCall: postMessage` and `TypeError: Cannot read properties of null (reading
  'postMessage')` — raised from `yt_player.html`'s own `setInterval`, where `WebClip.hostCall`
  hits a `contentWindow` that is `null` because the frame is not yet mounted or already gone.

Neither is caused by this change (relaxing an arity check cannot produce a null frame), but
the `null` dereference on a timer is a real if minor defect worth its own ticket.


deltaonlyone:
@review

deltaonlyone:
@review

yegordovganich (approved):

